PR #15, opened 2022-09-30
docs: README filepath changes

ethanholz:
Changed *.conf to *.tmuxtheme.


nekowinston:
Ah, my bad. Good catch, thanks!

PR #22, opened 2022-11-23
feat(window_tabs_enabled): Adding support for window tabs rather than directory tabs

### LINKED ISSUE #16: Use window titles for tabs instead of directory names
Hey, it's pretty common to use multiple tmux windows in the same directory. Currently, this config uses the name of the working directory to describe the tab and only shows the window title in a tab on the right. It seems like this should be the inverse, with the window titles describing the tabs and the directory name shown on the right.

Example from a project I'm working on:
CleanShot 2022-10-02 at 15 25 20@2x

Thank you for putting this config together!

rogeruiz:
This option, for now, defaults to `false`. This is just so it keeps the
initial design intention for the original design of the plugin. I may
flip it or change the name of the variable in future commits. This patch
is addressing catppuccin/tmux#16.

Alright, this closes #16 when it's merged into the default branch. If anyone would like to test this out locally, you _should_ be able to set a new remote to my GH fork and then change branches to `add-option-to-display-folders`.

The commands to do that in your local Git repository should be:

```sh
cd ~/.tmux/plugins/tmux
git remote add rogeruiz https://github.com/rogeruiz/tmux.git
git fetch --all
git checkout add-option-to-display-folders
```

Afterwards, you can follow the documentation in the `README.md` to turn on the feature. Remember to source your Tmux configuration file when you update it if you're in a session.

When you're done, you can remove the remote and go back to the default branch with the following:

```sh
cd ~/.tmux/plugins/tmux
git remote rm rogeruiz
git checkout main
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,4 @@
+# Let's not store the selected theme in version control. This file is used to
+# source theme variables locally in the main function rather than polluting the
+# global variable namespace.
+catppuccin-selected-theme.tmuxtheme
--- a/README.md
+++ b/README.md
@@ -46,6 +46,25 @@ set -g @catppuccin_flavour 'latte' # or frappe, macchiato, mocha
 1. Copy your desired theme's configuration contents into your Tmux config (usually stored at `~/.tmux.conf`)
 2. Reload Tmux by either restarting the session or reloading it with `tmux source-file ~/.tmux.conf`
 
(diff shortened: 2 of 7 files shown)


rogeruiz:
I don't see a `dev` branch here. so I'm merging this into `main` for now. But feel free to swap it out eventually.

PR #27, opened 2022-12-30
Add rounded icons in all places in the statusline

vaaleyard:
Never understood why keep the style on the left side different from the right side if it's already using the nerd icons.

This PR will make everything rounded.

**Preview:**
tabs_enabled

tabs_disabled

It's way prettier, in my opinion.

### DIFF
--- a/catppuccin.tmux
+++ b/catppuccin.tmux
@@ -65,13 +65,13 @@ main() {
   readonly wt_enabled
 
   # These variables are the defaults so that the setw and set calls are easier to parse.
-  readonly show_directory="#[fg=$thm_pink,bg=$thm_bg,nobold,nounderscore,noitalics]#[fg=$thm_bg,bg=$thm_pink,nobold,nounderscore,noitalics]  #[fg=$thm_fg,bg=$thm_gray] #{b:pane_current_path} #{?client_prefix,#[fg=$thm_red]"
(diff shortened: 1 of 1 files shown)


rogeruiz:
Thanks for this @vaaleyard. 

I'm realizing a pattern here in this PR, #32, & #30. Looks like we have lots of good contributors wanting the same thing. That thing is wanting more customization for the pill shapes and caps in the status line!

I will take a further look at this and see what we come up with.

I want to clarify this PR here, the customization in this PR about adding the cap to the left side for all elements in the status line instead of having the caps point to the center of the status line. Is this correct @vaaleyard?

vaaleyard:
I'm sorry, but I didn't understand your question. This PR just make all the statusline rounded instead of having only some things rounded (e.g. pane indicator).

rogeruiz:
Sorry if I was confusing @vaaleyard.

Looking closer at your images, I see the design you're going for a what I call a "pill" design. example: `( 1 | title )` versus what is currently on the default branch which is `| title | 1 )`. What the plugin currently uses is what I call a half-pill design. 

Let me know if you need any help refactoring your PR after #29 has been merged in.

the-spanish-guy:
Looks amazing 🤩 

I have an suggestion, u can add an conditional config ex: `set -g @catppuccin_left_separator "rounded" # or, bar, angled, etc `

nnhutan (approved):

89iuv:
Implemented in https://github.com/catppuccin/tmux/pull/57

PR #29, opened 2023-01-04
Improve performance

stasjok:
1. Don't create temporary files. It fixes an error for me when loading theme from a read-only filesystem (Nix store in my case).
2. Calling `tmux` for every variable is pretty slow. I changed it to aggregate all commands in a variable, then call it once.

Before:
```
Executed in   98.28 millis    fish           external
   usr time   53.09 millis    0.00 micros   53.09 millis
   sys time   20.00 millis  667.00 micros   19.33 millis
```

After:
```
Executed in   22.50 millis    fish           external
   usr time   14.36 millis  618.00 micros   13.74 millis
   sys time    5.53 millis  115.00 micros    5.41 millis
```

My first version printed commands in a function. Then I did `load_theme | tmux source-file -`, but looks like reading commands from stdin is pretty recent change in tmux (3.1): https://github.com/tmux/tmux/commit/5134666702ce972390f39e34bed9b60fe566263a. A second approach was `tmux source-file <(load_theme)`, but unfortunately it didn't work. Useful feature of using function for generating config lines would be that I also added command line switch, and user could `./catppuccin.tmux print >> ~/.config/tmux/tmux.conf`. But since it didn't work, it's not available.

If you don't like aggregation of commands, I can remove it, but I still want a fix for read-only filesystems.

### DIFF
--- a/catppuccin.tmux
+++ b/catppuccin.tmux
@@ -1,34 +1,37 @@
 #!/usr/bin/env bash
-PLUGIN_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+PLUGIN_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+get_tmux_option() {
+  local option value default
+  option="$1"
+  default="$2"
+  value="$(tmux show-option -gqv "$option")"
+
+  if [ -n "$value" ]; then
+    echo "$value"
+  else
+    echo "$default"
+  fi
+}
+
+set() {
+  local option=$1
+  local value=$2
+  tmux_commands+=(set-option -gq "$option" "$value" ";")
+}
+
+setw() {
+  local option=$1
+  local value=$2
+  tmux_commands+=(set-window-option -gq "$option" "$value" ";")
+}
 
 main() {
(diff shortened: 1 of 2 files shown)


megahirtz:
As a NixOS user I second the request to change how the the theme is loaded to allow it to work on read-only filesystems. That change alone would allow this plugin to be easily added to nixpkgs.

catppuccin.tmux
```diff
@@ -103,6 +110,8 @@ main() {
   #
   setw clock-mode-colour "${thm_blue}"
   setw mode-style "fg=${thm_pink} bg=${thm_black4} bold"
+
+  tmux "${tmux_commands[@]}"
```
rogeruiz, line 114: 🥳🥳 loving this optimization here.

rogeruiz (commented):
I'll prioritize this over the other PRs. Thanks @stasjok for this awesome work! I'll have to test this in macOS/Darwin myself to make sure everything is working too.

rogeruiz (approved):
lgtm 🌈

Sorry for taking so long to review this @stasjok but this is great work! Thank you for optimizing this plugin and removing the temporary file that broke compatibility with read-only file systems. 🎉

PR #30, opened 2023-01-05
feat(ui): make the separator configurable

89iuv:
Added a way to configure the separator. 
The default behavior is to use the rounded separator. 
Updated the documentation as well.

PR #32, opened 2023-01-22
add functionality for angled right status separators

rkarish:
New functionality for users who would prefer to have angled separators in the right status bar.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,14 @@ directory within the `status-right` and move the window names to the
 set -g @catppuccin_window_tabs_enabled on # or off to disable window_tabs
 ```
 
+##### Enable angled right status separators
+By default rounded right status separators are used, with `@catppuccin_right_status_separator`
+set to `angled` then angled separators will be used.
+
+```sh
+set -g @catppuccin_right_status_separator 'angled'
+```
+
 [style-guide]: https://github.com/catppuccin/catppuccin/blob/main/docs/style-guide.md
 
 ## 💝 Thanks to
--- a/catppuccin.tmux
+++ b/catppuccin.tmux
@@ -64,20 +64,34 @@ main() {
   wt_enabled="$(get-tmux-option "@catppuccin_window_tabs_enabled" "off")"
   readonly wt_enabled
 
+  # NOTE: Check for the value of @catppuccin_right_status_separator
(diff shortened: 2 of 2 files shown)


nekowinston:
@rogeruiz would appreciate if you can take a look at this. 🙂

rkarish:
Here is a preview of the angled separators

Screenshot 2023-01-22 at 3 21 09 AM

rogeruiz:
Awesome work on this @rkarish thanks for the additional customization here. 

I can work with you to get this in, but it looks like you missed a scenario where `@catppuccin_window_tabs_enabled` can be turned on and override the rounded/angled configuration entirely. 

I think this can be fixed this by having making this a bit simpler to implement. Let's add some variable place holders for the caps (rounded, angled). This way we don't need separate variables for rounded and angle. In the future we could combine your work so far with a custom cap to make it better for folks to customize the theme. I don't think we need to add this now but I'm hoping to communicate the intent here of having caps be a new variable which gets modified whenever your new variable (`@catppuccin_right_status_separator`) gets set to `rounded` or `angled`. 

Even if you don't want to make this a cap variable for easier reuse, the way this is currently implemented will cause the the first variable I mentioned to override the caps back to the hardcoded default of rounded even if this is turned on.

rkarish:
@rogeruiz thank you for the review! Since there are multiple PR's out there, such as #30, should I make changes to this PR per review or should we just consolidate this functionality into one PR? #30 has the functionality that I want and some more. The font pack that I use (Hacker Nerd Font Mono) makes the rounded separators not align correctly but the angled separators do. Also the angled separators match my Neovim setup 🙂. I'm in no rush to get this merged since I'm just using my fork for the tmux plugin currently. Let me know what you would like to do about moving this forward, I would be happy to make changes to this PR or I could help with #30 if that needs any changes.

rogeruiz:
Thanks @rkarish I've now merged in #29 so this branch will need some refactoring. If #30 fills your intentions with this PR, then let's close this and leverage #30. If @89iuv can't rebase their work onto the new changes, you can take a try and getting the branch up to date and discuss those changes there.

I'll wait about 4 weeks or so to proceed if I don't hear back. Thanks for understanding.

rkarish:
Hi @rogeruiz lets leverage #30. I will close this. I will chime in over there to see if I can be of any help to get those conflicts resolved and rebase the branch.

PR #35, opened 2023-03-02
Add hostname and date time as an option to configure

steinbrueckri:
This patch adds the functionality to show the hostname and the date and time in the status bar.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,18 @@ directory within the `status-right` and move the window names to the
 set -g @catppuccin_window_tabs_enabled on # or off to disable window_tabs
 ```
 
+##### Enable hostname in status bar
+
+```sh
+set -g @catppuccin_hostname_enabled on
+```
+
+##### Enable time and date in status bar
+
+```sh
+set -g @catppuccin_datetime_enabled on
+```
+
 [style-guide]: https://github.com/catppuccin/catppuccin/blob/main/docs/style-guide.md
 
 ## 💝 Thanks to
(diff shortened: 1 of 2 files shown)


rogeruiz:
Hi @steinbrueckri, I've done some testing on this. And I like the direction here. I don't seem to have the icon you're using for `hostname` in my NerdFont. Could you include a screenshot here so I can see what it is? 

Also, I want to find a more compatible icon here. Are you using the latest NerdFont compilation or some other icon?

steinbrueckri:
@rogeruiz This is what I use in my Alacritty config ...

```yaml
font:
  normal:
    family: Hack Nerd Font Mono

  bold:
    family: Hack Nerd Font Mono

  italic:
    family: Hack Nerd Font Mono

  bold_italic:
    family: Hack Nerd Font Mono
```

It looks like this ...
Bildschirm­foto 2023-03-03 um 09 21 58

Im not a designer so if you have other options on the colors or the icons feel free to change them. 🔰

ercpereda:
I test it, and it is working correctly.
My comments are that the date and time should be managed independently, 
and also have an option to specify the format could be good.

rogeruiz:
Okay after attempting to build my own nerd font to get that font to render, I think a compromise here would be to allow the configuration to override the icon for these. What do you think? 

Also while looking into this PR further, I realize that I have some changes in my own personal fork that adds custom functions and custom icons to the left and the right. I wonder if we should incorporate your changes into that concept and have folks be able to add their own custom functionality with prescribing that they need to have the `hostname` and `date/time` on their status line. What do you think @steinbrueckri? https://github.com/rogeruiz/tmux/blob/personal-tmux-line/catppuccin.tmux#L198-L224

steinbrueckri:
To be honest, I think your solution is better. 👌
But we should add some config examples with screenshots how it can look.

89iuv:
fixed in https://github.com/catppuccin/tmux/pull/38

Issue #41, opened 2023-05-02
There is no way to toggle `show directory`


Diego-Llanes:
I know that this is minor but it really bugs me having to seen the directory icon constantly changing in the bottom right, if there is a solution to this please let me know.  From what I can see the only way to change this would be by modifying the source files.

89iuv:
Feature has been implemented in: https://github.com/catppuccin/tmux/pull/57

PR #43, opened 2023-05-03
Make all segments optinal

ercpereda:
Solves #41


89iuv:
Implemented in https://github.com/catppuccin/tmux/issues/59

PR #57, opened 2023-08-08
Feature/refactor in modules

89iuv:
This PR intends to break the configuration file into modules so that it is easier to maintain.
The benefits are:
- modular approach to defining the status components
- increased customizability

Note. This PR has breaking changes. Because of the refactor some of the previous configuration option are not available anymore or have been replaced.

# List of removed options

## Status line themes
```sh
set -g @catppuccin_pill_theme_enabled on
set -g @catppuccin_powerline_theme_enabled on
set -g @catppuccin_powerline_icons_theme_enabled on
set -g @catppuccin_no_patched_fonts_theme_enabled on
```

This have been removed in favor of letting the user configure the look and feel of tmux.
If you are having issues with the new settings please reply here https://github.com/catppuccin/tmux/issues/58 and i will do my best to help.

## Enable window tabs
```sh
set -g @catppuccin_window_tabs_enabled on # or off to disable window_tabs
```

This has been removed because the same can be achieved with the new settings.

## Configure separator 
```sh
set -g @catppuccin_l_left_separator ""
set -g @catppuccin_l_right_separator ""
set -g @catppuccin_r_left_separator ""
set -g @catppuccin_r_right_separator ""
```

This settings have been replaced with:
```sh
set -g @catppuccin_window_separator_left ""
set -g @catppuccin_window_separator_right ""
set -g @catppuccin_status_separator_left ""
set -g @catppuccin_status_separator_right ""
```

## DateTime, User and Host
```sh
set -g @catppuccin_date_time "%Y-%m-%d %H:%M"
set -g @catppuccin_user "on"
set -g @catppuccin_host "on"
```
This modules have been reworked
To enable a module add it to the lists of modules on the position that you want:
```sh
set -g @catppuccin_status_modules" "application session user host date_time"
 ```
To configure a module's icon, text or color: (except for session module)
```sh
set -g @catppuccin_<module_name>_[icon | text | color]" "value"
```


catppuccin.tmux
```diff
@@ -31,6 +31,160 @@ setw() {
...
+  local custom_icon_window_current="$(get_tmux_option "@catppuccin_icon_window_current" "󰖯")"
+  local custom_icon_window_zoom="$(get_tmux_option "@catppuccin_icon_window_zoom" "󰁌")"
+  local custom_icon_window_mark="$(get_tmux_option "@catppuccin_icon_window_mark" "󰃀")"
+  local custom_icon_window_silent="$(get_tmux_option "@catppuccin_icon_window_silent" "󰂛")"
+  local custom_icon_window_activity="$(get_tmux_option "@catppuccin_icon_window_activity" "󰖲")"
+  local custom_icon_window_bell="$(get_tmux_option "@catppuccin_icon_window_bell" "󰂞")"
+
+  if [[ $window_status_icon_enable == "yes" ]]
+  then
+    local show_window_status="#(printf '%%s\n' '#F' | sed \"s/*/${custom_icon_window_current}/\" | sed \"s/-/${custom_icon_window_last}/\" | sed \"s/#/${custom_icon_window_activity}/\" | sed \"s/#//g\"| sed \"s/~/${custom_icon_window_silent}/\" | sed \"s/!/${custom_icon_window_bell}/\" | sed \"s/M/${custom_icon_window_mark}/\" | sed \"s/Z/${custom_icon_window_zoom}/\")"
```
Stonks3141, line 47: This can be done with a single sed invocation like this: `sed 's/../../; s/../../'`

Also note that sed will only substitute the first match in a line by default, you have to write `s/../../g` to substitute all matches in a line. I’m not sure which one you want here, just letting you know.

catppuccin.tmux
```diff
@@ -31,6 +31,160 @@ setw() {
...
+  fi
+
+  if [[ $window_status_icon_enable == "no" ]]
```
Stonks3141, line 50: `[[` and `==` will only work in Bash, not in POSIX shells. Several Linux distros such as Alpine and all BSDs do not come with Bash. Additionally, if the variable is empty for some reason, an error will occur. It should be quoted to avoid this. This can be written more portably as `[ "$foo" = "bar" ]`.

catppuccin.tmux
```diff
@@ -31,6 +31,160 @@ setw() {
...
+  echo $window_format
+}
+
+build_status_module() {
+  local index="$1"
+  local icon="$2"
+  local color="$3"
+  local text="$4"
+
+  if [[ $index -eq 0 || $status_connect_separator == "no" ]]
```
Stonks3141, line 132: With single brackets, the and and or operators have to go outside the brackets, like so: `[ .. ] && [ .. ]`

catppuccin.tmux
```diff
@@ -31,6 +31,160 @@ setw() {
...
+}
+
+load_modules() {
+  local loaded_modules
+
+  local modules_path=$1
+  local modules_list=$2
+
+  local modules_array
+  read -a modules_array <<< "$modules_list"
```
Stonks3141, line 168: Arrays and `<<<` are Bashisms, the portable equivalent would be to remove this variable and replace the for loop below with `echo "$modules_list" | while read -r module_name`.

catppuccin.tmux
```diff
@@ -31,6 +31,160 @@ setw() {
...
+  local module_name
+  for module_name in ${modules_array[@]}
+  do
+    local module_path=$modules_path/$module_name.sh
+    source $module_path
+
+    if [[ 0 -eq $? ]]
+    then
+      loaded_modules="$loaded_modules$( show_$module_name $module_index )"
+      module_index=$module_index+1
```
Stonks3141, line 180: In portable shell, `$((` must be used for arithmetic, like this: `module_index=$((module_index + 1))`.

catppuccin.tmux
```diff
@@ -31,6 +31,160 @@ setw() {
...
+    fi
+
+  done
+
+  echo $loaded_modules
```
Stonks3141, line 185: This should maybe be quoted.

catppuccin.tmux
```diff
@@ -65,141 +219,29 @@ main() {
...
-  fi
+  local status_modules="$(get_tmux_option "@catppuccin_status_modules" "application session")"
+  local loaded_modules=$( load_modules "$PLUGIN_DIR/status" "$status_modules")
 
   set status-left ""
-  set status-right "${right_column1}${right_column2}"
-
-  setw window-status-format "${window_status_format}"
-  setw window-status-current-format "${window_status_current_format}"
+  set status-right "${loaded_modules}"
```
Stonks3141, line 244: No need for curly braces here.

Stonks3141 (commented):
Hi! I’ve left some comments on the use of Bash-specific shell features, since AFAIK tmux doesn’t use Bash on non-GNU/Linux systems like Alpine (which uses busybox) or other Unixes.

catppuccin.tmux
89iuv: @Stonks3141 I am scratching my head here trying to implement this suggestion. The problem is that i need to update 2 variables from inside of the loop. From my understanding using a "|" creates a subshell where the while loop runs and the variables outside of the subshell will not get updated.

Any ideas on how to move forward with this?

catppuccin.tmux
89iuv: I am assuming here the the characters i am substituting will only appear once in the line.

89iuv:
@Stonks3141 I implemented all the suggestions.

PR #64, opened 2023-08-11
Remove sed dependencies

vdbe:
A drawback for the custom status icons is that the window_marked_flag was only added in [tmux 3.1](https://github.com/tmux/tmux/blob/3.1/CHANGES#L120C26-L120C26).

For context Debian 11/Bullseye is the first debian release with tmux 3.1 (or newer):
 - [Debian 10/Buster - tmux](https://packages.debian.org/buster/tmux)
 - [Debian 11/Bullseye - tmux](https://packages.debian.org/bullseye/tmux)

### DIFF
--- a/catppuccin.tmux
+++ b/catppuccin.tmux
@@ -44,7 +44,8 @@ build_window_icon() {
 
   if [ "$window_status_icon_enable" = "yes" ]
   then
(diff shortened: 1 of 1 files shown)


89iuv:
I killed tmux-server and after that the theme was not loaded. 

It looks like something with seed function that got replaced around the line:
# NOTE: Pulling in the selected theme by the theme that's being set as local

Does is work for you after a tmux server restart?

89iuv:
Did a fix here: https://github.com/catppuccin/tmux/commit/10d8db8b88294e094007b85e593a5c29f6b823db

PR #71, opened 2023-08-29
feat(status): allow modules in left_status

lowstar:
Added possibility of adding modules also to the left of the status line. 

Example:

```sh
set -g @catppuccin_status_modules "host"
set -g @catppuccin_left_status_modules "session"
```
Screenshot 2023-08-29 at 09 18 25

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@
    1. [Config 1](#config-1)
    2. [Config 2](#config-2)
    3. [Config 3](#config-3)
-   
+
 ## Themes
 
 - 🌻 [Latte](./catppuccin-latte.tmuxtheme)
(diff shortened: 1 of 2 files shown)


89iuv:
Good work. 

It would be nice, if you have the time, to refactor so that we have @catppuccin_status_modules_left and @catppuccin_status_modules_right in order to be more clear on where the loaded modules go.

lowstar:
> It would be nice, if you have the time, to refactor so that we have @catppuccin_status_modules_left and @catppuccin_status_modules_right in order to be more clear on where the loaded modules go.

Good idea. PR updated with new commit.

89iuv:
Thank you

Issue #80, opened 2023-09-10
Background color for default vs current window


bendavis78:
I find the background color for the default vs current window a bit confusing. It seems that the current window should have the lighter bg, while the default should have the same bg as the whole line. Instead it's inversed, so if two windows are opened it looks like the non-current window is current.

89iuv:
Indeed it was confusing for me at first. I did not want to mess with the original vision of the author and change the default.
The solution here would be to make the background configurable for default and current window.

formigarafa:
I had the same initial feeling but, wondered about configuration options too.
but to be honest, I think it had been a mistake to set the active window to be the same as bg color. It just doesn't make sense. 
Even your separator icons won't appear if you do so. 
Why would one go to the lengths to make the icons configurable if they are not visible?
<img width="378" alt="image" src="https://github.com/catppuccin/tmux/assets/160616/da1b539f-25ba-47b8-b9a0-243df46f06b5">

I am no specialist here but just gave a bash to change `background` from `$thm_bg` to `$thm_black4` which they provide and the stars seem to align for me:
<img width="355" alt="image" src="https://github.com/catppuccin/tmux/assets/160616/89078ebe-e7ae-4ad6-8911-8583bf9e3786">

sgoudham:
Closing as https://github.com/catppuccin/tmux/pull/94 was merged, making this user configurable.

PR #94, opened 2023-11-08
Feature: Window BG Color

thezacharytaylor:
Adding a variable to allow the modification of the current window while otherwise defaulting to the original orange. Likewise was done for the default window, but with the gray. The variable names used were kept honest to the existing variables.

Addresses:
#80

Relevant options added:

`set -g @catppuccin_window_current_color "#DDB6F2"`
`set -g @catppuccin_window_default_background "#333333"`

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,6 +141,15 @@ Values:
 - all - the entire window part will have the same color
 - none - the entire window part will have no color
 
+#### Override the window default colors:
+```sh
+set -g @catppuccin_window_default_text "color" # text color
+set -g @catppuccin_window_default_background "color"
+```
+
+Values:
+- color - a hexadecimal color value
+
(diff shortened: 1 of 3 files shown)


89iuv:
Can you please add the fallowing configs as well:
```sh
set -g @catppuccin_window_default_color "#DDB6F2"
set -g @catppuccin_window_current_background "#333333"
```

This way we will have a way to overwrite both the color and the background for both the current and default window.
ex:
```sh
set -g @catppuccin_window_default_color "#DDB6F2"
set -g @catppuccin_window_default_background "#333333"

set -g @catppuccin_window_current_background "#333333"
set -g @catppuccin_window_current_color "#DDB6F2"
```

Please also update the documentation.

PR #95, opened 2023-11-25
Use highlighted current window text background

formigarafa:
Changes current window text background color so it looks selected and it's separator icons don't disappear in the background:

Before:
image

After:
image


89iuv:
I think this approach https://github.com/catppuccin/tmux/pull/94 is better.

PR #106, opened 2024-01-03
feat: Configure pane (active) border styles

alewis001:


Em1lT:
Really good addition ❤️

PR #117, opened 2024-01-15
feat(weather): add weather module for xamut/tmux-weather

edman:
Weather module
Requirements
This module depends on https://github.com/xamut/tmux-weather.

Install
The preferred way to install tmux-weather is using TPM.

Configure
Load tmux-weather after you load catppuccin.

set -g @plugin 'catppuccin/tmux'
...
set -g @plugin 'xamut/tmux-weather'

Add the weather module to the status modules list.

set -g @catppuccin_status_modules" "... weather ..."

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@
    5. [Customizing modules](#customizing-modules)
    6. [Battery module](#battery-module)
    7. [CPU module](#CPU-module)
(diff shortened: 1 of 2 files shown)


edman:
> Can you please update documentation also?

Done. I think there was no documentation when I uploaded my PR. Maybe I missed it.

PR #137, opened 2024-02-17
docs(README.md) use correct option for default window color

henrisota:
I noticed that the wrong option was being referred to on the documentation when setting the default window color.


sgoudham (approved):
Thanks!

PR #148, opened 2024-03-02
feat(config): allow status bar to be disabled by default

connordeckers:
This adds a new config entry for setting the default status bar visibility on load, rather than forcing it to be on by default.


connordeckers:
I like to have my status bar off initially, and have a keybind to toggle it. I was running `set -g status "off"` after TPM loaded the plugins, but it led to flickering on load/reload. This means that it can be disabled (though continues to default to "on"), providing configurability without regression :)

I've also updated the readme to include the new configuration item.

PR #151, opened 2024-03-06
Minor Spelling errors corrected  in  README.md

cortiz:
Minor Spelling errors corrected 

In the 
### Weather Module  change 

`The preferred way to install tmux-cpu is using` to  `The preferred way to install tmux-weather is using`

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,8 +45,8 @@
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


sgoudham (approved):
I'm not sure all of these are necessarily spelling _errors_ but thanks!

PR #153, opened 2024-03-07
docs(custom): fix outdated module name

### LINKED ISSUE #131: Little Error In README.md
Hey there,

I think there is a small error in `README.md`. I was using one of the example configs, and it took me a while to understand why no matter what I did, the modules on the right would not accurately reflect what was specified in the config. I thought I would leave an issue so others don't have the same trouble. Eventually I realized the issue was that I need to use this:

`set -g @catppuccin_status_modules "application session user host date_time"`

But all the examples in `README.md` use this:

`set -g @catppuccin_status_modules_right "application session user host date_time"`

Awesome repo, thank you for making this!

sgoudham:
I'm _really_ confused at how the `custom` folder was seemingly gitignored but also kept git history... I'm assuming its supposed to be in git history and have deleted the `.gitignore` as it was the only entry.

Closes #131


vdbe:
@sgoudham `custom` was added to `.gitignore` so no custom user modules made it upstream by accident.
A better way to ignore it would be to ignore `custom` except current files.
Here is PR for it #154

PR #154, opened 2024-03-07
fix(gitignore): add gitignore

vdbe:
Ignore custom (with the exception of documentation/examples) so no custom user modules make it upstream


sgoudham (approved):
Thanks for informing me on why it was originally included! LGTM

PR #157, opened 2024-03-10
feat: add nowplaying module

xwjdsh:
The module is based on tmux plugin https://github.com/xwjdsh/tmux-nowplaying.


vdbe:
I suggest that you keep this as custom module since it is a macos only 4 days old plugin.

PR #159, opened 2024-03-14
ci: add testing & shellcheck workflows

vdbe:
Add a test workflow to see if everything works still works.

This will just test if nothing errors **not** if the all the options are correctly set.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,35 @@
+---
+name: Tests
+permissions:
+  contents: read
+on:
+  pull_request:
+    paths-ignore:
+      - "*.md"
+      - "assets/**"
+  push:
+    paths-ignore:
+      - "*.md"
+      - "assets/**"
+    branches:
+      - main
+  workflow_dispatch:
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  ubuntu:
+    name: Test
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Install tmux
+        run: sudo apt-get update && sudo apt-get install -y tmux=3.2a-4ubuntu0.2
+      - uses: actions/checkout@v4
+      - name: Config and Plugin
+        shell: bash
+        run: |
+          bash --version
+          tmux -V
+          ./.github/scripts/test.bash ./catppuccin.tmux
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -33,3 +33,21 @@ jobs:
           bash --version
           tmux -V
           ./.github/scripts/test.bash ./catppuccin.tmux
+
+  old-bash:
+    name: "Old Bash"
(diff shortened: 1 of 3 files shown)


vdbe:
Example run: https://github.com/vdbe/catppuccin-tmux/pull/1

.github/workflows/test.yml
```diff
@@ -0,0 +1,53 @@
...
+  cancel-in-progress: true
+
+jobs:
+  ubuntu:
+    name: Test
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Install tmux
+        run: sudo apt-get update && sudo apt-get install -y tmux
```
sgoudham, line 28: We should pin the version of tmux here so that the version of tmux doesn't implicitly change between runs.

.github/workflows/test.yml
```diff
@@ -0,0 +1,53 @@
...
+    name: "Old bash"
+    runs-on: ubuntu-latest
+    container:
+      image: bash:3.2.57
+    steps:
+      - uses: actions/checkout@v4
+      - name: config & plugin (old bash)
+        shell: bash
+        run: |
+          # No tmux available in this container
```
sgoudham, line 47: I might be misunderstanding the point of this but if tmux isn't available in this container, why would we be running the test script?

sgoudham (changes requested):

.github/workflows/test.yml
vdbe: To check if this bash version does not error on "new" syntax `catppuccin.tmux`

image

commit: 11a92b95fdf9c47035e8d1c58db6e92a6a5e5b4f

.github/workflows/test.yml
vdbe: Is `apt-get install tmux=3.2a-4ubuntu0.2` fixed enough?

sgoudham: Does a later version such as https://github.com/tmux/tmux/releases/tag/3.4 exist in apt? I just wanted to make sure it was pinned to at least one of the versions in the releases

sgoudham:
> Example run: [vdbe#1](https://github.com/vdbe/catppuccin-tmux/pull/1)

This run has failed with the following error:
```
./catppuccin.tmux:369:6: error: Double quote array expansions to avoid re-splitting elements. [SC2068]
```

We should fix that error in this PR

.github/workflows/test.yml
vdbe: `ubuntu-latest` action runner is [22.04/jammy](https://packages.ubuntu.com/jammy/tmux) for which 3.2a the latest is. Compile from source?

sgoudham: Ah no worries, 3.2a is fine :+1: 
Don't think its worth it for a simple test

vdbe:
That is a PR to test the workflows, made commit to test if catppuccin.tmux was checked by the shellcheck action.
Will not merged.

sgoudham (approved):

Issue #160, opened 2024-03-15
Add tmux-clima as an additional module for weather


xafarr:
Clima (Esperanto for weather) is a simple plugin that displays the local weather conditions and temperature in your Tmux status line based on **OpenWeatherMap API**. For more information, please have a look at the project. Following is the link to the project.

[tmux-clima](https://github.com/vascomfnunes/tmux-clima)

PR #161, opened 2024-03-15
Add tmux-clima as another module for weather

xafarr:
#160

### DIFF
--- a/README.md
+++ b/README.md
@@ -373,15 +373,22 @@ Add the cpu module to the status modules list.
 set -g @catppuccin_status_modules_right "... cpu ..."
 ```
 
-### Weather module
+### Weather modules
 
-#### Requirements
+There are a couple of modules that can be used to retrieve weather data.
+
+- [tmux-weather](https://github.com/xamut/tmux-weather)
+- [tmux-clima](https://github.com/vascomfnunes/tmux-clima)
+
+#### [tmux-weather](https://github.com/xamut/tmux-weather)
+
+##### Requirements
 This module depends on [tmux-weather](https://github.com/xamut/tmux-weather).
 
-#### Install
+##### Install
 The preferred way to install tmux-weather is using [TPM](https://github.com/tmux-plugins/tpm).
 
-#### Configure
+##### Configure
 Load tmux-weather after you load catppuccin.
 ```sh
 set -g @plugin 'catppuccin/tmux'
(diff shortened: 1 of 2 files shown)


vdbe:
What does this plugin provide that the the current weather module doesn't?
It seems to also use OpenWeatherMap so the result are the same I guess (and you need your own api key).

Before merging this it needs documentation in the README.

xafarr:
It does the same thing but in a different way. Also, [tmux-weather](https://github.com/xamut/tmux-weather) is pretty inactive. And [tmux-clima](https://github.com/vascomfnunes/tmux-clima) can be an extra option for people to use with **catppuccin/tmux**. Updated the PR with README update. Thanks.

README.md
```diff
@@ -363,15 +363,22 @@ Add the cpu module to the status modules list.
 set -g @catppuccin_status_modules_right "... cpu ..."
 ```
 
-### Weather module
+### Weather modules
```
vdbe, line 366: Link under `## Content` needs to be updated

README.md
```diff
@@ -363,15 +363,22 @@ Add the cpu module to the status modules list.
...
 
-#### Requirements
+There are a couple of modules that can be used to retrieve weather data.
+
+- [tmux-weather](https://github.com/xamut/tmux-weather)
+- [tmux-clima](https://github.com/vascomfnunes/tmux-clima)
+
+#### [tmux-weather](https://github.com/xamut/tmux-weather)
```
vdbe, line 373:
```suggestion
#### tmux-weather
```
Already linked 3 lines below

README.md
```diff
@@ -384,6 +391,27 @@ Add the weather module to the status modules list.
 set -g @catppuccin_status_modules_right "... weather ..."
 ```
 
+#### [tmux-clima](https://github.com/vascomfnunes/tmux-clima)
```
vdbe, line 394:
```suggestion
#### https://github.com/vascomfnunes/tmux-clima
```

vdbe (changes requested):

vdbe:
It currently does not do anything extra except set prefix+W.
I will leave this open in case this changes and for further discussion/requests.

xafarr:
I have been using **tmux-clima** for almost 2 years and contribute to it as well. I also like **catppuccin** project and use it in all terminal programs. I raised this PR for people like me to have an option to use these together. I used to use **tmux-weather** but was constantly having problems with *wttr.in*. That was the reason I moved to this plugin. *OpenWeatherMap API* is pretty stable and gives free 1000 calls per day.

vdbe (approved):

PR #162, opened 2024-03-16
feat(window-list): allow customization of window list position

andresatierf:
Allow customization of the window list position with `@catppuccin_status_justify`


vdbe:
Duplicate of #145

Issue #134, opened 2024-02-15
How to change status bar color?


LaurensDSchunk:
## Issue
The tmux status bar is not updating its color. I want it to look like it does in this repo's images. I am using alacritty as my terminal. Perhaps just a way to change the color of the status bar?

## Screenshot
<img width="798" alt="Screenshot 2024-02-15 at 8 06 25 AM" src="https://github.com/catppuccin/tmux/assets/101758235/936643e8-9995-42fa-b9ce-32a754228399">

## Desired Effect (Darkest status bar)

<img width="100" alt="Screenshot 2024-02-16 at 11 46 10 AM" src="https://github.com/catppuccin/tmux/assets/101758235/eb6a016d-a24e-4fb0-be08-72faf662eee4">

## .tmux.conf
```
bind r
bind r source-file ~/.tmux.conf

set -g prefix C-s

setw -g mode-keys vi
bind-key h select-pane -L
bind-key j select-pane -D
bind-key k select-pane -U
bind-key l select-pane -R
set-option -g remain-on-exit on

set -sg escape-time 10
set-option -g status-position top
set -g mouse on

set -g @plugin 'catppuccin/tmux'
set -g @plugin 'tmux-plugins/tpm'
set -g @plugin 'tmux-plugins/tmux-sensible'
set -g @plugin 'christoomey/vim-tmux-navigator'

# Fix Colors
set -g default-terminal "screen-256color"
set -as terminal-features ",xterm-256color:RGB"

set -g @catppuccin_flavour 'mocha' # latte or frappe, macchiato, mocha

set -g @catppuccin_window_left_separator ""
set -g @catppuccin_window_right_separator " "
set -g @catppuccin_window_middle_separator " █"
set -g @catppuccin_window_number_position "right"

set -g @catppuccin_window_default_fill "number"
set -g @catppuccin_window_default_text "#W"

set -g @catppuccin_window_current_fill "number"
set -g @catppuccin_window_current_text "#W"

set -g @catppuccin_status_modules_right "directory session"
set -g @catppuccin_status_left_separator  " "
set -g @catppuccin_status_right_separator ""
set -g @catppuccin_status_right_separator_inverse "no"
set -g @catppuccin_status_fill "icon"
set -g @catppuccin_status_connect_separator "no"

set -g @catppuccin_directory_text "#{pane_current_path}"

run '~/.tmux/plugins/tpm/tpm'
```

LaurensDSchunk:
[...]

Issue #87, opened 2023-10-15
Add option to specify separator colors


druskus20:
![image](https://github.com/catppuccin/tmux/assets/43417195/a168759f-2679-4675-8948-9572b170b06c)
I want to change the background of the entire Tmux bar, however, the separators have a different color. I can't find a way to set the separator color.

```
# To set the bar color, use: 
set -g status-bg "#181825"
```

druskus20:
Actually, just as I opened this, I figured out a way:

```
set -g @catppuccin_status_left_separator "#[fg=#ffffff,bg=#ffffff,bold,underscore,italics]█"
set -g @catppuccin_status_right_separator "#[fg=#ff0000,bg=#ff0000,bold,underscore,italics]█"
```

It's a bit of a hack but it works

druskus20:
I've almost managed to do it, with a hack. 

```
# Set background for the status bar
set -g @catppuccin_window_right_separator "#[fg=#313244,bg=#181825]█ "
set -g @catppuccin_status_left_separator "█"
set -g @catppuccin_status_left_separator "#[bg=#181825] █" 
set -g @catppuccin_window_current_fill "number"

# other stuff
set -g @catppuccin_status_left_separator_inverse "no"
set -g @catppuccin_status_fill "all"
set -g @catppuccin_status_modules_right "directory session user host"
set -g @catppuccin_directory_text "#{pane_current_path}"
set -g @catppuccin_window_current_text "#W"
```

![image](https://github.com/catppuccin/tmux/assets/43417195/edb6ce6f-0bbd-4276-a6a3-429fd4c2f196)

**The only problem, is that window middle separator color is not configurable:**

https://github.com/catppuccin/tmux/blob/89ad057ebd47a3052d55591c2dcab31be3825a49/catppuccin.tmux#L63
https://github.com/catppuccin/tmux/blob/89ad057ebd47a3052d55591c2dcab31be3825a49/window/window_current_format.sh#L4

Which results in: 

![image](https://github.com/catppuccin/tmux/assets/43417195/2cc9c62d-59ca-4840-85c8-20a844ef31bd)

Could we, either add an option to tweak this, or just options for each separator?

Issue #84, opened 2023-09-14
Add documentation/examples on how to overwrite background and enable transparency


git-girl:
hey firstly thanks for more catppuccin :sunflower: 

i was trying to setup the plugin and wanted to have it have a transparent background. without the plugin in the `tmux.conf` and with the `set-option -g status-style bg=default` it works fine. but once i add the plugin the status bar has a black background. 

i also tried putting the `set-option status-style` before and after the plugin load, and the different versions of the option (allthough the one in the config is the correct one for v3.3a that i'm on).  i also tried settting the alacritty TERM var to `xterm-256color` as described in [the tmux faq](https://github.com/tmux/tmux/wiki/FAQ) 

did i miss an option i have to set for this?

<details>
<summary>this is my tmux.conf</summary>

``` conf
set -g @plugin 'tmux-plugins/tpm'
# actually pretty minimal https://github.com/tmux-plugins/tmux-sensible
set -g @plugin 'tmux-plugins/tmux-sensible'

set -g @plugin 'sainnhe/tmux-fzf'
# yank on y
set -g @plugin 'tmux-plugins/tmux-yank'

set -g @plugin 'catppuccin/tmux'

set -g mouse on

set -g base-index 1
set -g pane-base-index 1 
set-window-option -g pane-base-index 1
set-option -g renumber-windows on

# from https://github.com/aserowy/tmux.nvim
is_vim="ps -o state= -o comm= -t '#{pane_tty}' | grep -iqE '^[^TXZ ]+ +(\\S+\\/)?g?\.?(view|n?vim?x?)(-wrapped)?(diff)?$'"

bind-key -n 'C-h' if-shell "$is_vim" 'send-keys C-h' { if -F '#{pane_at_left}' '' 'select-pane -L' }
bind-key -n 'C-j' if-shell "$is_vim" 'send-keys C-j' { if -F '#{pane_at_bottom}' '' 'select-pane -D' }
bind-key -n 'C-k' if-shell "$is_vim" 'send-keys C-k' { if -F '#{pane_at_top}' '' 'select-pane -U' }
bind-key -n 'C-l' if-shell "$is_vim" 'send-keys C-l' { if -F '#{pane_at_right}' '' 'select-pane -R' }

bind-key -T copy-mode-vi 'C-h' if -F '#{pane_at_left}' '' 'select-pane -L'
bind-key -T copy-mode-vi 'C-j' if -F '#{pane_at_bottom}' '' 'select-pane -D'
[...]

PR #170, opened 2024-03-23
feat(status, window): add option to make overwrite status background …

89iuv:
This PR adds 2 new options

### Set the window separator
```sh
set -g @catppuccin_window_separator ""
```

#### Override the default status bar background color
```sh
set -g @catppuccin_status_background "theme"
```
This will overwrite the status bar background:
- "theme" will use the color from the selected theme
- "default" will make the status bar transparent
- use hex color codes for other colors

This should fix issues:
- https://github.com/catppuccin/tmux/issues/134
- https://github.com/catppuccin/tmux/issues/87
- https://github.com/catppuccin/tmux/issues/84

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,25 +18,25 @@
 ## Content
 
 1. [Themes](#themes)
-2. [Installation](#installation)
-3. [Overview](#overview)
-4. [Configuration options](#configuration-options)
+1. [Installation](#installation)
+1. [Overview](#overview)
+1. [Configuration options](#configuration-options)
    1. [Window](#window)
-   2. [Window default](#window-default)
-   3. [Window current](#window-current)
-   4. [Status](#status)
-   4. [Pane](#pane)
-   5. [Customizing modules](#customizing-modules)
-   6. [Battery module](#battery-module)
-   7. [CPU module](#CPU-module)
-   8. [Weather modules](#weather-modules)
-   9. [Load module](#load-module)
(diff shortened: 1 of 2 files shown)


vdbe (approved):

Gaurav-Gosain:
Thank you so much for this change! Not sure if this is the right place to ask this but do you know of a way to get rid of these backgrounds and make it completely transparent?

image

vdbe:
`$thm_bg` needs to be replaced with `$status_background` inside [`built_status_module`](https://github.com/catppuccin/tmux/blob/809361f5c2a7b5eb2fa75d981dcb8e2f9454800b/catppuccin.tmux#L201-L265)

Gaurav-Gosain:
So I managed to get it working for my local setup by doing the following:

image

I use catppuccin-mocha so I modified `~/.tmux/plugins/tmux/catppuccin-mocha.tmuxtheme` and set the `thm_bg` variable to default.

```
thm_bg="default"
```

This gets most of the job done and should be plenty for most. However after doing this, the background of the icons changed to a light gray which didn't look good with my setup so I went ahead and modified the [`fg for show_icon`](https://github.com/catppuccin/tmux/blob/809361f5c2a7b5eb2fa75d981dcb8e2f9454800b/catppuccin.tmux#L211) to use `$thm_gray` instead of `$thm_bg`.

Hope this helps someone :)

PR #173, opened 2024-03-24
fix: status background for modules

vdbe:


vdbe:
Some text/icon used `thm_bg` as foreground.
Maybe it's a better solution to add `local color_bg=$5`/`local color_text=$5` instead of reusing the status background.
image

vdbe:
@Gaurav-Gosain What do you think?

Gaurav-Gosain:
@vdbe thank you so much for this, I will test it out, but just having a quick glance, it looks like a good solution to the problem!

89iuv:
I am getting this result: 
image

I am using: 
```sh
set -g @catppuccin_status_background "red"
```
This changes only the status background leaving the other part of tmux unchanged.

vdbe:
#170 only changed the status bar color, not colors that matched the status-bar by using `$thm_bg` as background (or foreground in some cases)

With the config
```sh
# set -g @catppuccin_status_background "#ff0000"
set -g @catppuccin_status_background "red"
```
I get 
- Main
image
- This branch
image

89iuv:
I understand the issue now. 

There is a disconnect between the status color which is red and the separator background which is $thm_bg.

It would be better to somehow make that part transparent so that it blends into the background, but until then the proposed solution also works.

vdbe:
It's not just the `build_status_module` `build_window_format` has the same issue.
Also just replace `thm_bg` with `status_background` for now?

Gaurav-Gosain:
So I managed to get it working for my local setup by doing the following:

image

I use catppuccin-mocha so I modified `~/.tmux/plugins/tmux/catppuccin-mocha.tmuxtheme` and set the `thm_bg` variable to default.

```
thm_bg="default"
```

This gets most of the job done and should be plenty for most. However after doing this, the background of the icons changed to a light gray which didn't look good with my setup so I went ahead and modified the [`fg for show_icon`](https://github.com/catppuccin/tmux/blob/809361f5c2a7b5eb2fa75d981dcb8e2f9454800b/catppuccin.tmux#L211) to use `$thm_gray` instead of `$thm_bg`.

vdbe:
Replaced by #175

PR #175, opened 2024-03-27
feat: improve transparent background

89iuv:
This PR improves the handling of the background color and transparency.

In order for this to work the fallowing setting had to be extended to windows settings:
```sh
set -g @catppuccin_status_connect_separator "yes"
```

Note:
- you need to restart the tmux server in order for the new background settings to take effect.

Example:
Solid:
image

Transparent:
image

Todo:
- test this on other configs


vdbe (commented):
Only tested example configs.

Config 2 breaks with this change
It does however use `set -g @catppuccin_status_connect_separator "no"`

With `set -g @catppuccin_status_connect_separator "yes"`
image

With `set -g @catppuccin_status_connect_separator "no"`
image

Current main (`set -g @catppuccin_status_connect_separator "no`)
image

I guess this replaces #174 ?

89iuv:
@vdbe Good find with config 2. I have some ideas on how to fix that one.

> I guess this replaces https://github.com/catppuccin/tmux/pull/174 ?

This PR is not related to https://github.com/catppuccin/tmux/pull/174. I think we should merge  https://github.com/catppuccin/tmux/pull/174 and announce the breaking change as a pinned issue.

vdbe:
> This PR is not related to https://github.com/catppuccin/tmux/pull/174. I think we should merge  https://github.com/catppuccin/tmux/pull/174 and announce the breaking change as a pinned issue.

Sorry meant #173

89iuv:
Yes, i think this should replace https://github.com/catppuccin/tmux/pull/173.

89iuv:
Run into a problem here:

The objective of the option ```@catppuccin_status_right_separator_inverse``` is to swap between the foreground and the background color for the status right separator.
The problem is that while the background color can be set to transparent, the foreground color defaults to white.

This creates the situation where we need to choose between having this option or having a transparent background.

My proposal is to remove this option as i do not think it is very popular and by doing this we can fully support having a transparent background (requested multiple times) without any other issue.

@vdbe what do you think?

vdbe:
Removing `@catppuccin_status_right_separator_inverse` does seem like the best option

vdbe (approved):
Just updating the example configs

- Replace config 2?
- Remove `catppuccin_status_right_separator_inverse` from config 3

89iuv:
Good catch. I forgot about updating the config example. Will work on this next.

catppuccin.tmux
```diff
@@ -89,14 +85,11 @@ main() {
   local window_status_separator=$(get_tmux_option "@catppuccin_window_separator" "")
   setw window-status-separator "$window_status_separator"
 
-  setw window-status-activity-style "fg=${thm_fg},bg=${thm_bg},none"
-  setw window-status-style "fg=${thm_fg},bg=${thm_bg},none"
-
```
vdbe, line 94: Just making sure this was not removed by accident.

89iuv: This is needed because in tmux once you set the bg to a solid color, it can't be set to transparent afterwards.

It is better to have the builder decide on the colors and the set them in tmux only once.

PR #182, opened 2024-04-02
refactor(whiskers): title

vdbe:
A few notes
- Move to standard names or hardcode `thm_...` (https://github.com/catppuccin/community/blob/main/rfc/CAT-0003-Whiskers.md#variables)
  - What does thm stand for?
- Change theme format so we can just source it instead of parsing?
- Move themes to a separate folder
  - Currently only 3 themes (the main flavours) but can be more in the future (accents)

https://github.com/catppuccin/community/discussions/15


vdbe:
See #207

Issue #177, opened 2024-03-29
Feature request: Window connect seperator 


ian-vu:
Hey 👋👋, I am looking for the ability to connect the window separators similar to how the current status separators work. I am not sure if it is currently supported.

Currently there is support for the option `catppuccin_status_connect_separator "yes"` which allows for no gaps between the modules.  I am looking for the same effect but for the windows.

Hopefully my screenshot better expresses what I am looking for. 

![CleanShot 2024-03-29 at 20 22 56@2x](https://github.com/catppuccin/tmux/assets/42108124/a2a714b1-3f87-4005-aa4c-680af45c4ca0)

MykhailoSukhostavets:
Just created PR, for now problem that I see is with most right separator. To fix this dracula tmux uses fixed icon that renders before all windows. 
In this example from dracula icon shows current session (on screenshot only one window, can be confusing)
<img width="1466" alt="image" src="https://github.com/catppuccin/tmux/assets/28440361/a14025ab-2e6b-4125-a447-2446dca2e1ac">

If we are interested in the same logic, I can create try to implement it

ian-vu:
That looks awesome @MykhailoSukhostavets thanks for your work! 🙏 Perfectly what I was trying to express

kjnsn:
This is supported by the tmux option `window-status-separator`. Closing

PR #184, opened 2024-04-03
Window connect separator 

sukhostavets:
Issue: #177

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ set -g @catppuccin_icon_window_bell "󰂞 "
 
 #### Set the window connect separator:
 ```sh
-set -g @catppuccin_window_connect_separator "yes"
+set -g @catppuccin_window_connect_separator "no"
 ```
 Values:
 - yes - windows will join with a small gape between them
--- a/README.md
+++ b/README.md
@@ -145,6 +145,14 @@ set -g @catppuccin_icon_window_activity "󱅫 "
 set -g @catppuccin_icon_window_bell "󰂞 "
 ```
 
+#### Set the window connect separator:
+```sh
+set -g @catppuccin_window_connect_separator "yes"
+```
+Values:
+- yes - windows will join with a small gape between them
+- no - windows will not join
+
 ### Window default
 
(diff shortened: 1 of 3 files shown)


vdbe:
There seems to be an issue with the background between the windows

`set -g @catppuccin_status_background "#ff0000"` with config:
- provided in #177 
Screenshot from 2024-04-08 10-57-19

- [config 2](https://github.com/catppuccin/tmux#config-2)
Screenshot from 2024-04-08 10-56-45

- [config 3](https://github.com/catppuccin/tmux#config-3)
Screenshot from 2024-04-08 10-55-35

vdbe:
I think you need to set `window-status-style` bg (and prbly also the active style) like done with `status-style` to get default working.

https://github.com/catppuccin/tmux/blob/5ed4e8a6a20c928688da268dfcdf460ac9c3cb49/catppuccin.tmux#L56-L65

MykhailoSukhostavets:
Thank you, I will test it later.
(Also I still need to update logic when we fill number)

MykhailoSukhostavets:
> I think you need to set `window-status-style` bg (and prbly also the active style) like done with `status-style` to get default working.

Didn't work or I did smth wrong, I tried to play with logic like that:
```
      set-window-option -g window-status-style fg="#ff0000"
      set-window-option -g window-status-style bg="#ff0000"
      set-window-option -g window-status-current-style fg="#ff0000"
      set-window-option -g window-status-current-style bg="#ff0000"
```

Other than "default" should work fine
image

vdbe:
Is there _any_ possibility to reduce the line count of `build_window_format` it was bad before but this a 100 line switch statement with 4 almost identical cases/branches

Some patterns (for window_connect_seperator blocks):
- bg consists 1 color and default (always at the same place)
- first 2 show_left always `$thm_bg`

vdbe:
https://github.com/catppuccin/tmux/blob/33c4fd62df96b39cec2b2c08bc341c88d9392761/builder/window_builder.sh#L52

bg needs to be `default` for connected and `$background` for not connected

Issue #197, opened 2024-04-23
Feature Request: Specific 'separator' configuration for each window type


mklbravo:
### Is there an existing issue outlining your improvement?

- [X] I have searched the existing issues and my improvement has not been raised yet.

### What would you like to see added and/or changed?

I want to configure my tmux using the following window separator configuration:

```conf
set -g @catppuccin_window_left_separator "█"
set -g @catppuccin_window_middle_separator "█ "
set -g @catppuccin_window_number_position "left"
set -g @catppuccin_window_right_separator "█"

# Different fill configuration for each window type
set -g @catppuccin_window_current_fill "all"
set -g @catppuccin_window_default_fill "number"
```

Using this configuration, separator will look on on `window_default_fill` but is looks weird when using `window_current_fill`. 
<img width="179" alt="Screenshot 2024-04-23 at 23 09 10" src="https://github.com/catppuccin/tmux/assets/2145612/8b2a8ad0-2364-4c48-aeb6-ea6e6f27e4e2">

It will be nice to be able to configure different separators for each window type, so that I can change it the same way I do with the fill configuration.

My idea will be to add the following options:
```conf
# Window global configuration. 
set -g @catppuccin_window_left_separator "█"
set -g @catppuccin_window_middle_separator "█ "
set -g @catppuccin_window_right_separator "█"

# Current type window configuration. If set, it will override the global configuration
set -g @catppuccin_window_current_left_separator "█"
set -g @catppuccin_window_current_middle_separator " 󰿟 "
set -g @catppuccin_window_current_right_separator "█"

# Default type window configuration. If set, it will override the global configuration
set -g @catppuccin_window_default_left_separator "█"
set -g @catppuccin_window_default_middle_separator " 󰿟 "
set -g @catppuccin_window_default_right_separator "█"
```

I don't know if this could be of help to anyone else.

mklbravo:
Thank you @vdbe!

PR #205, opened 2024-05-09
fix(window_builder): number position left

vdbe:
Middle separator color was inverted when number position left.

Before:
Screenshot from 2024-05-09 18-56-26

After:
Screenshot from 2024-05-09 18-56-13


vdbe:
This fixes a part of the issue from #197

PR #207, opened 2024-05-10
refactor(theme): use whiskers

vdbe:


themes/latte/latte.tmuxtheme
vdbe, line 1: With the "new" colors:
image

templates/catppuccin-flavor.tmuxtheme.tera
```diff
@@ -0,0 +1,26 @@
+---
+whiskers:
+  version: "2.1.0"
+  matrix:
+    - flavor
+  filename: "themes/{{flavor.identifier}}/{{flavor.identifier}}.tmuxtheme"
```
sgoudham, line 6: Wouldn't it make more sense to have a directory structure like:

```
themes/
├─ catppuccin_latte.tmuxtheme
├─ catppuccin_frappe.tmuxtheme
├─ catppuccin_macchiato.tmuxtheme
├─ catppuccin_mocha.tmuxtheme
```

Feels like the flatter version is nicer, is there a practical reason for nesting it further?

themes/latte/latte.tmuxtheme
sgoudham: Looks good!

sgoudham (changes requested):

templates/catppuccin-flavor.tmuxtheme.tera
sgoudham, line 1: Our informal convention thus far for the file name is `<port name>.tera`, which in this case would be `tmux.tera`. I think it'd be nice to update it but I'm happy to approve this PR even without the rename :+1:

templates/catppuccin-flavor.tmuxtheme.tera
vdbe: Was from when the matrix included accents

sgoudham (approved):

PR #216, opened 2024-05-12
Fix/test old bash

### LINKED ISSUE #215: Test bash version 3.2.57 properly with tmux instead of mocking it
### Is there an existing issue outlining your improvement?

- [X] I have searched the existing issues and my improvement has not been raised yet.

### What would you like to see added and/or changed?

We can use docker image `bash:3.2.57-alpine3.19` and add tmux to comes with tmux 3.4

vdbe:
Actually use tmux instead of bypassing it to test compatibility with old bash version 3.2.57.
This will prevent accidents like `color_values+=("${val:1:-1}")` in #208 from going unnoticed  before merging.

Closes #215

### DIFF
--- a/.github/scripts/test.bash
+++ b/.github/scripts/test.bash
@@ -59,7 +59,7 @@ main() {
     exit 1
   else
     echo "Running plugin: $plugin" >&2
-    output=$(tmux run-shell "$plugin" 2>&1)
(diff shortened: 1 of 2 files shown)


vdbe:
Example failed test: https://github.com/vdbe/catppuccin-tmux/pull/5

Issue #236, opened 2024-06-10
Support for mode indicator via tmux-mode-indicator


rbpatt2019:
### Is there an existing issue outlining your improvement?

- [X] I have searched the existing issues and my improvement has not been raised yet.

### What would you like to see added and/or changed?

Two in a row, I know, but I was missing these features :P Support for displaying the current mode via [tmux-mode-indicator](https://github.com/MunifTanjim/tmux-mode-indicator) would be great. Are you open to contributions?

vdbe:
Pull requests are welcome see [custom](https://github.com/catppuccin/tmux/tree/main/custom) on how to create a module. Don't forget to set the fg/bg for the plugin with it's style options, I don't know how those colors will react with the separators so maybe only use the fg and match bg with the module.

rbpatt2019:
@vdbe - thanks for the advice! I'll get started in the next couple of days

PR #237, opened 2024-06-13
Add suport for tmux-mode-indicator

rbpatt2019:
Adds support for [tmux-mode-indicator](https://github.com/MunifTanjim/tmux-mode-indicator).
Classically, this plugin changes its background colour by mode. There is no single
`bg` setting in the plugin, but rather a separate one for each mode. To make the
catppuccin `bg` change dynamically as well, we would then need to re-implement the
logic for detecting the mode. To avoid this complexity, this PR sets all the
`bg` colours to the catppuccin background colour and only changes the text.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,8 @@
 <h3 align="center">
-	<img src="https://raw.githubusercontent.com/catppuccin/catppuccin/main/assets/logos/exports/1544x1544_circle.png" width="100" alt="Logo"/><br/>
-	<img src="https://raw.githubusercontent.com/catppuccin/catppuccin/main/assets/misc/transparent.png" height="30" width="0px"/>
-	Catppuccin for <a href="https://github.com/tmux/tmux">Tmux</a>
-	<img src="https://raw.githubusercontent.com/catppuccin/catppuccin/main/assets/misc/transparent.png" height="30" width="0px"/>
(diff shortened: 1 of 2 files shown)


rbpatt2019:
Close #236.

README.md
vdbe, line 1: Only change segments related to this PR (`### Mode Indicator module`)

status/mode_indicator.sh
vdbe, line 1: Some options to present the modes with colors:

- change to fg color to thm_blue/yellow/red/cyan (if this  remains readable)

- create the equivalent of https://github.com/MunifTanjim/tmux-mode-indicator/blob/7027903adca37c54cb8f5fa99fc113b11c23c2c4/mode_indicator.tmux#L46 but just for background color and use that as `color` (don't think you need a rewrite to make it dynamic)
- Same as above but set `$thm_bg` _should_ only effect this module (don't really like this)

vdbe (changes requested):

vdbe:
Feels like it would make more sense to extend the session module with color indicators for copy and sync, but then you loose the text representation.

Starting to lean more to making this a builtin module instead of requiring a plugin if extending session is not enough, since  re-implementing in an opinionated way would be as much or less work as configuring it in an opinionated way.

https://github.com/catppuccin/tmux/blob/e7cd05facc22c18fbab3f718d8ecd6494538fe38/status/session.sh#L6

PR #240, opened 2024-06-26
perf: batch tmux show-options

vdbe:
Batching tmux show-options gives a significant start uptime decrease of up to 3 under linux (tested with bath bash-5.2.26 and 3.2.57).

~~On the bash version used by macos this however seems to have a negative impact. This was test in docker with as image `bash:3.2.57-alpine3.19`. The negative impact maybe from docker/alpine since in `bash:5.1-alpine3.19` it also had a negative impact but less.
Can anyone on macos see what the real effect is?~~

~~I think the the complexity increase is worth it if the "macos"/old bash/alpine hit is because of alpine, but would love to get more feedback on this.~~
Effect was docker/alpine build bash-3.2.57 and it works great.

Tested on 10 user configs (from issues) with and without batching and batching was always an improvement.

With defaults in the array is most of the time slower but did not spend any time in looking for improvements since the with defaults is a bit more user friendly.

- Example config 1
Screenshot from 2024-07-14 16-32-33

- Example config 2
Screenshot from 2024-07-14 16-35-22

- Example config 3
Screenshot from 2024-07-14 16-38-49

### DIFF
--- a/utils/tmux_utils.sh
+++ b/utils/tmux_utils.sh
@@ -57,78 +57,78 @@ setw() {
 }
 
 get_tmux_batch_option() {
-	local option default value
-	option="$1"
+  local option default value
+  option="$1"
   default="$2"
 
-	for option_index in "${!tmux_batch_options[@]}"; do
-		IFS=":" read -r read_option read_value <<<"${tmux_batch_options[$option_index]}"
-		if [[ "$read_option" == "$option" ]]; then
-			echo "$read_value"
+  for option_index in "${!tmux_batch_options[@]}"; do
+    IFS=":" read -r read_option read_value <<<"${tmux_batch_options[$option_index]}"
+    if [[ "$read_option" == "$option" ]]; then
+      echo "$read_value"
       return
-		fi
-	done
+    fi
+  done
 
   echo "$default"
 }
 
 get_interpolated_tmux_batch_option() {
-	local option default value
-	option="$1"
+  local option default value
+  option="$1"
   default="$2"
 
-	for option_index in "${!tmux_batch_options[@]}"; do
-		IFS=":" read -r read_option read_value <<<"${tmux_batch_options[$option_index]}"
-		if [[ "$read_option" == "$option" ]]; then
+  for option_index in "${!tmux_batch_options[@]}"; do
+    IFS=":" read -r read_option read_value <<<"${tmux_batch_options[$option_index]}"
+    if [[ "$read_option" == "$option" ]]; then
       do_color_interpolation "$read_value"
       return
-		fi
-	done
+    fi
+  done
 
   echo "$default"
 }
 
 add_tmux_batch_option() {
-	local option
-	option="$1"
+  local option
+  option="$1"
 
-	tmux_batch_options_commands+=("show-option -gq $option ;")
+  tmux_batch_options_commands+=("show-option -gq $option ;")
 }
 
 set_tmux_batch_option() {
-	local option value
-	option="$1"
-	value="$2"
+  local option value
+  option="$1"
+  value="$2"
 
-	for option_index in "${!tmux_batch_options[@]}"; do
-		read -d ':' -r read_option <<<"${tmux_batch_options[$option_index]}"
-		if [[ "$read_option" == "$option" ]]; then
-			tmux_batch_options["$option_index"]="$option:$value"
-			return
-		fi
-	done
+  # NOTE: don't check for duplicates just append
+  # for option_index in "${!tmux_batch_options[@]}"; do
+  #   read -d ':' -r read_option <<<"${tmux_batch_options[$option_index]}"
+  #   if [[ "$read_option" == "$option" ]]; then
+  #     tmux_batch_options["$option_index"]="$option:$value"
+  #     return
+  #   fi
+  # done
 
   tmux_batch_options+=("$option:$value")
-	tmux_batch_options_commands+=("show-option -gq $option ;")
 }
 
 run_tmux_batch_commands() {
-	local temp
+  local temp
 
   # shellcheck disable=SC2048,SC2086
-	while IFS=' ' read -r option value; do
-		if [ -n "$value" ]; then
-			if [ "$value" = "null" ]; then
-				set_tmux_batch_option "$option" ""
-			else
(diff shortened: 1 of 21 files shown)


vdbe:
1.28+ with ./window and ./pane.
./status is harder since it's not always loaded

Screenshot from 2024-06-26 13-25-20

vdbe:
I have been using this PR for 2 weeks now and everything seems to work.

Would love to have the performance increases/decreases from other people/platforms so we can make a better decision on including this or not.

The _bash_ script below will 'benchmark' this PR and the main branch using [hyperfine](https://github.com/sharkdp/hyperfine) if installed and `time` if not. 
It would be really helpful if you could run this and comment the output/result (and if it's linux/macos/wsl/...).

```bash
#!/usr/bin/env bash

# Clone both repositories/branches
git clone --depth=1  https://github.com/catppuccin/tmux.git catppuccin-tmux
git clone --depth=1 --branch perf/batch-tmux-show https://github.com/vdbe/catppuccin-tmux.git catppuccin-tmux-batch

# Print bash and tmux version
echo "bash: ${BASH_VERSION}"
echo "tmux: $(tmux -V)"

# Benchmark
if command -v hyperfine 
then
  # hyperfine is installed
  hyperfine -N ./catppuccin-tmux/catppuccin.tmux ./catppuccin-tmux-batch/catppuccin.tmux
else
  # hyperfine is not installed just using sleep
  echo ./catppuccin-tmux/catppuccin.tmux
  time ./catppuccin-tmux/catppuccin.tmux
  echo ./catppuccin-tmux-batch/catppuccin.tmux
  time ./catppuccin-tmux-batch/catppuccin.tmux
fi
```
[benchmark.bash.txt](https://github.com/user-attachments/files/16142573/benchmark.bash.txt)


<details>
  <summary>Example outputs</summary>

```
bash: 5.2.26(1)-release
tmux: tmux 3.4
./catppuccin-tmux/catppuccin.tmux

real	0m0.450s
user	0m0.231s
sys	0m0.183s
./catppuccin-tmux-batch/catppuccin.tmux

real	0m0.263s
user	0m0.144s
sys	0m0.110s
```

```
bash: 5.2.26(1)-release
tmux: tmux 3.4
/nix/store/9hv3swqri0zg5iqaw0v4mxsycxc5vgbb-hyperfine-1.18.0/bin/hyperfine
Benchmark 1: ./catppuccin-tmux/catppuccin.tmux
  Time (mean ± σ):     399.6 ms ±   3.0 ms    [User: 221.4 ms, System: 157.6 ms]
  Range (min … max):   396.6 ms … 406.4 ms    10 runs

Benchmark 2: ./catppuccin-tmux-batch/catppuccin.tmux
  Time (mean ± σ):     244.1 ms ±   2.3 ms    [User: 131.4 ms, System: 106.2 ms]
  Range (min … max):   241.7 ms … 248.5 ms    12 runs

Summary
  ./catppuccin-tmux-batch/catppuccin.tmux ran
    1.64 ± 0.02 times faster than ./catppuccin-tmux/catppuccin.tmux
```

</details>

utils/tmux_utils.sh
```diff
@@ -54,3 +54,75 @@ setw() {
...
+		if [[ "$read_option" == "$option" ]]; then
+			tmux_batch_options["$option_index"]="$option:$value"
+      # shellcheck disable=SC2034
+			tmux_batch_commands["$option_index"]="show-option -gq $option ;"
+			return
+		fi
+	done
+
+	set_tmux_batch_option_default "$option" "$value"
+}
```
vdbe, line 109: Don't keep track of default's in the array.
This will reduce the array size to not defaults but force `get_tmux_options` to walk the entire list before falling back to the default.

in `./cattpuccin.tmux` we currently have 34 `@catppuccin_...` options of which most will not be set so total compares should go down? This might be worth exploring.
I expect current implementation to be _worst_ case.

uncenter:
On macOS:

```
bash: 5.2.26(1)-release
tmux: tmux 3.4
/etc/profiles/per-user/uncenter/bin/hyperfine
Benchmark 1: ./catppuccin-tmux/catppuccin.tmux
  Time (mean ± σ):     230.6 ms ±  43.4 ms    [User: 87.4 ms, System: 90.1 ms]
  Range (min … max):   209.6 ms … 343.2 ms    10 runs

  Warning: The first benchmarking run for this command was significantly slower than the rest (343.2 ms). This could be caused by (filesystem) caches that were not filled until after the first run. You should consider using the '--warmup' option to fill those caches before the actual benchmark. Alternatively, use the '--prepare' option to clear the caches before each timing run.

Benchmark 2: ./catppuccin-tmux-batch/catppuccin.tmux
  Time (mean ± σ):      99.8 ms ±  38.7 ms    [User: 38.6 ms, System: 41.8 ms]
  Range (min … max):    88.4 ms … 228.7 ms    13 runs

  Warning: The first benchmarking run for this command was significantly slower than the rest (228.7 ms). This could be caused by (filesystem) caches that were not filled until after the first run. You should consider using the '--warmup' option to fill those caches before the actual benchmark. Alternatively, use the '--prepare' option to clear the caches before each timing run.

Summary
  ./catppuccin-tmux-batch/catppuccin.tmux ran
    2.31 ± 1.00 times faster than ./catppuccin-tmux/catppuccin.tmux
```

utils/tmux_utils.sh
vdbe: Some quick math:
```python
>>> total_options=34
>>> compare_count_current=sum(list(range(1,total_options+1)))
>>> compare_count_current
595
>>> options_count_break_even=compare_count_current/total_options
>>> options_count_break_even
17.5
>>> 17*total_options
578
```

in formulas:

n = # options
a = # not default options
x = # compares with default in [] = n + (n-1) + ... + 2 + 1 = (n*(n+1))/2
y = worst case # compares without default in [] = a * n

break even for n => a(n):
  x(n) = y(a, n) <=> (n*(n+1))/2 = a*n <=> (n+1)/2 = a
  a(n) = (n+1)/2

  a(34) = (34+1)/2 = 35/2 = 17.5

In other words if _50% or less_ of the options are _not default_ it should be faster to not track default in the list/array

Issue #245, opened 2024-07-12
Application module is showing window name instead


Mikilio:
### Is there an existing issue outlining your problem?

- [X] I have searched the existing issues and they do not solve my problem.

### Describe your problem.

The [default](https://github.com/catppuccin/tmux/blob/ff6dacb12f3291abe116bd63430a2b5842a1e65c/status/application.sh#L7) for the application module is the window name, not the actual application name. I assume this is not intended behavior because it leads to confusion when people actually rename their windows to something that makes sense for them.

### Paste your configuration.

Doesn't really matter

### What tmux version are you seeing the issue on?

tmux 3.4

### Any additional comments?

Will fix myself.

Current workaround is:
```
set -g @catppuccin_directory_text "#{pane_current_command}"
```

vdbe:
That is because the window name is the applications name.
![image](https://github.com/user-attachments/assets/df6b2c45-01c4-4c35-b0f8-e696c7b4ee85)
bottom is current default top is your suggested change.

I don't see how that workaround improves/fixes anything, maybe I am missing something?

Mikilio:
> That is because the window name is the applications name.

That was what I was trying to say: Users are free to rename the window, so that doesn't hold for everyone.

> bottom is current default top is your suggested change.

Wow! That is surprising! Definitely not how it looks on my system.

With these settings:
```

      set -g @catppuccin_window_left_separator ""
      set -g @catppuccin_window_right_separator " "
      set -g @catppuccin_window_middle_separator " █"
      set -g @catppuccin_window_number_position "right"

      set -g @catppuccin_window_default_fill "number"
      set -g @catppuccin_window_default_text "#W"

      set -g @catppuccin_window_current_fill "number"
      set -g @catppuccin_window_current_text "#W#{?window_zoomed_flag, -  ,}"

      set -g @catppuccin_status_modules_right "application user host session"
      set -g @catppuccin_application_text "#{pane_current_command}"
      set -g @catppuccin_status_left_separator  " "
      set -g @catppuccin_status_right_separator ""
      set -g @catppuccin_status_fill "icon"
      set -g @catppuccin_status_connect_separator "no"

      set -g @catppuccin_directory_text "#{pane_current_path}"
```

I get this
![image](https://github.com/user-attachments/assets/ab717d79-c235-48ef-a4b2-9d2e096bece8)
(I have changed my window name to code)

vdbe:
[...]

PR #246, opened 2024-07-12
fix: better default for application.sh

Mikilio:
Solves #245


vdbe:
Thank you did not know about `pane_current_command`, should RTFM myself.

PR #247, opened 2024-07-13
feat(menu): add menu style options

cluther:
To allow `#{thm_x}` styling of menus.

- @catppuccin_menu_style (`menu-style`)
- @catppuccin_menu_selected_style (`menu-selected-style`)
- @catppuccin_menu_border_style (`menu-border-style`)


vdbe (changes requested):
This change has some unintended side effects:
- does not work with transparency
  This can be fixed by using `bg=default` instead of `bg=${thm_gray}` (maybe also only setting fg have not tested that).
- disabled options are no longer grayed out.
  Don't know how to fix this only other menu option I can find is `menu-border-lines` (https://github.com/tmux/tmux/blob/master/options-table.c#L332-L366).
  Maybe default to `""` and only set when not `""`?

Current:
Screenshot from 2024-07-13 19-08-15


This PR:
Screenshot from 2024-07-13 19-15-55

cluther:
Thanks, @vdbe. I'll test those defaults more carefully, and update the pull request.

cluther:
@vdbe: Switching to `bg=default` as you suggested works. I can't reproduce the problem with disabled menu options. They do appear disable for me.

This is with tmux 3.4 installed from nixos-unstable, and configured with catppuccin/nix.
[catppuccin-tmux-menu-transparency.webm](https://github.com/user-attachments/assets/7159ea5c-9f65-4cbf-9c95-bfefbb54de28)

vdbe:
Weird can you try it with just `set -g mouse on` in the config?

image

image

image

https://github.com/vdbe/catppuccin-tmux-dev

cluther:
I had mouse enabled. Thanks for the flake. That helped me reproduce the problem. I still don't totally understand why the disabled items were losing styling under that basic config, and not my own, but it was solely related to changing the `menu-style` value. By explicitly setting it to `default` the problem goes away. So I set `menu-border-style` that way as well just to be safe. Setting `menu-selected-style` to `default` doesn't work, so I made that as close to the tmux default of `bg=yellow,fg=black` as I could.

Based on testing with your flake as you showed, this looks good to me now.

catppuccin.tmux
```diff
@@ -85,6 +85,15 @@ main() {
   set message-style "fg=${thm_cyan},bg=${message_background},align=centre"
   set message-command-style "fg=${thm_cyan},bg=${message_background},align=centre"
 
+  # menu
+  local menu_style menu_selected_style menu_border_style
+  menu_style=$(get_interpolated_tmux_option "@catppuccin_menu_style" "default")
+  menu_selected_style=$(get_interpolated_tmux_option "@catppuccin_menu_selected_style" "fg=${thm_black4},bg=${thm_yellow}")
```
vdbe, line 91: Why change to black4? I think gray was more readable.
image
What do you think?

README.md
```diff
@@ -208,6 +208,25 @@ set -g @catppuccin_pane_border_style "fg=#{thm_gray}" # Use a value compatible w
...
+#### Set the menu selected style:
+
+```sh
+set -g @catppuccin_menu_selected_style "fg=#{thm_black4},bg=#{thm_yellow}" # Use a value compatible with the standard tmux `menu-selected-style`
+```
+
+### Set the menu border style:
+
+```sh
+set -g @catppuccin_menu_border_style "fg=#{thm_fg},bg=default" # Use a value compatible with the standard tmux `menu-border-style`
```
vdbe, line 228: Update `style` and `border_style` to `default` and fg for `selected` to `gray depending on what is used.

vdbe (changes requested):

catppuccin.tmux
vdbe: just occurred to me I only tested 1 flavor.

vdbe: image
image
image
image


gray looks in all better I think.

cluther: I prefer gray too. I only changed to black4 because it seemed to match the main branch more closely.

cluther: Gray it is!

vdbe (approved):

PR #264, opened 2024-08-03
Improve the handling of separation between window title and status

j-xella:
1. Before, it was always separated by space, now the separator can be changed or removed via global option
2. Now, the separator is only added if there is an actual status to display. Works for custom separator icons only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,6 +133,12 @@ Values:
 - yes - this will replace the windows status text with icons
 - no - this will keep the windows status in text format
 
+#### Set window status icon/text separator:
+```sh
+set -g @catppuccin_window_status_separator " "
+```
+The character to use for separating window text and status.
+
 #### Override windows status icons
 ```sh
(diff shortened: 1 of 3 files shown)


vdbe:
This is a decision made by the theme, see #168

j-xella:
OK, understood, the theme wants window status indicators always separated by spaces. 

But this PR also has another part to it, a bug fix IMHO. Similar to what #265 does to status modules.

Let's say user sets active window icon to null. Then the icon is not shown, but the separator space is added nevertheless to the window info. IMO, there should be no separator if status icon is removed, correct?

Would this PR be considered if only that part is left?

vdbe:
if you can find a could way to remove the space if it's empty yes.

Best way I could come up with is using this `echo "[${icon:+ $icon}]"`
image

j-xella:
Please see #270

Issue #85, opened 2023-10-02
Remove status icon and background fill


alexchantastic:
Is there a way to completely remove the status icons and background fill? So far, I've only been able to remove the icons, but a small piece of the background fill remains (the colorful rectangles in the screenshot below).

My guess is that the icon module actually contains a space?

![image](https://github.com/catppuccin/tmux/assets/604167/b240d389-3fcb-4971-aa3c-3317a32d223e)

Here's my tmux config:

```
set -g @catppuccin_flavour "mocha"
set -g @catppuccin_status_left_separator "null"
set -g @catppuccin_status_modules_right "application session date_time"
set -g @catppuccin_session_icon "null"
set -g @catppuccin_application_icon "null"
set -g @catppuccin_date_time_icon "null"
```

89iuv:
Indeed there is a space between the icon and the text and when you remove the icon the space is still there.

vdbe:
You could change the color of the module to match the background `set -g @catppuccin_<module>_color "#{thm_gray}" (depending on your `@catppuccin_status_fill` this should be `default` not `#{thm_gray}").

Will leave an extra space.

![image](https://github.com/catppuccin/tmux/assets/44153531/9c13f319-d6d3-4416-bfb4-d3bb562d4646)

j-xella:
Will it possibly be fixed by #265 ?

vdbe:
Was indeed fixed by #265.

PR #265, opened 2024-08-03
Don't add icon separator if there is no icon for a status module

j-xella:
When rendering status modules, there is a separator between the icon and the module text. Before, it was redundantly shown even when the module icon was disabled by the user.


vdbe (changes requested):
A better way to do this is before the echo check if an icons is empty and clear `$show_icon` if so

```bash
if [ -z "$icon] ; then
  show_icon=""
fi
```

j-xella:
Reimplemented, please have a look

j-xella:
It could possibly fix #85

vdbe (approved):

PR #270, opened 2024-08-04
fix(window): don't prepend separator before empty status icon

j-xella:
Only works for catppuccin_window_status_icon_enable = yes

### DIFF
--- a/builder/window_builder.sh
+++ b/builder/window_builder.sh
@@ -95,6 +95,12 @@ build_window_format() {
   echo "$final_window_format"
 }
 
+prepend_separator() {
+  local field="$1"
+
+  echo "${field:+ $field}"
+}
+
 build_window_icon() {
   # Only update `show_window_status` if it's not empty
   # this module is ran twice once for current and once for default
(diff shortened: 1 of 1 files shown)


builder/window_builder.sh
```diff
@@ -109,18 +116,18 @@ build_window_icon() {
...
+    show_window_status+="#{?window_active,$(prepend_separator "${custom_icon_window_current}"),}"
+    show_window_status+="#{?window_last_flag,$(prepend_separator "${custom_icon_window_last}"),}"
+    show_window_status+="#{?window_marked_flag,$(prepend_separator "${custom_icon_window_mark}"),}"
+    show_window_status+="#{?window_zoomed_flag,$(prepend_separator "${custom_icon_window_zoom}"),}"
 
   fi
 
   if [ "$window_status_icon_enable" = "no" ]; then
-    local show_window_status=" #F"
+    local show_window_status=$(prepend_separator "#F")
```
vdbe, line 130: This _always_ needs to have a space.

```suggestion
    local show_window_status=" #F"
```

builder/window_builder.sh
```diff
@@ -92,6 +92,13 @@ build_window_format() {
   echo "$final_window_format"
 }
 
+prepend_separator() {
+  local field="$1"
+  local separator="${2:- }"
+
+  echo "${field:+$separator$field}"
+}
+
```
vdbe, line 101: This function can just be 1 line, but would prefer to just skip the function and inline it everywhere.

vdbe (changes requested):
Merge conflict and some other changes.

builder/window_builder.sh
j-xella: > This _always_ needs to have a space.

The reason I changed it here is because I wanted consistency of code between custom icons and text display of window status. And it is going to **always** have space, because tmux patterns are not yet expanded at that point, so it is just "#F" which is not empty and hence the space separator is always prepended.

builder/window_builder.sh
j-xella: > This function can just be 1 line, 

Yes, it could be. Something like this:

```bash
echo "${1:+${2:- }$1}"
```

But this way the function seems clearer to me and easier to understand its purpose. I was actually wondering if it could be reused in other places too? But for that it would have to be in some generic script, I was not sure where is the best to put it...

> but would prefer to just skip the function and inline it everywhere.

I was thinking about it too. It would save a couple of lines of code, but the code would be less maintainable as the result, IMO. 

Having an additional function is probably also a tiiiiny bit slower, but not sure anyone would ever notice.

builder/window_builder.sh
vdbe: > The reason I changed it here is because I wanted consistency of code

Consistency is not adding dead/useless code, especially not dead/useless code that makes it _harder_ to read.

builder/window_builder.sh
vdbe: Why do you need that `$2` it's never provided it's _always_ going to be a space.

vdbe (changes requested):

builder/window_builder.sh
j-xella: In this case - yes. But I was wondering if this function could be reused elsewhere, where the separator could be configurable ...

j-xella:
> Merge conflict 

resolved

builder/window_builder.sh
j-xella: Restored

builder/window_builder.sh
vdbe: You can always add it in the future without changing current behavior.

j-xella:
OK, I updated the PR not to have conflicts with the latest state of the code. 

Also, I removed the second parameter from the prepend_separator function.

Hope that is enough ...

vdbe (approved):

PR #277, opened 2024-08-05
refactor: `@catppuccin_flavour` -> `@catppuccin_flavor`

vdbe:
Most catppuccin ports have adopted the flavor spelling.

`@catppuccin_flavour` is still available but no longer documented and `@catppuccin_flavor` takes priority.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ set -g @plugin 'tmux-plugins/tpm'
 3. (Optional) Set your preferred flavor, it defaults to `"mocha"`:
 
 ```bash
-set -g @catppuccin_flavour 'mocha' # latte,frappe, macchiato or mocha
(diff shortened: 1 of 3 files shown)


sgoudham:
Can we output a deprecation warning for `flavour` before removing it in v1?

vdbe:
> Can we output a deprecation warning for `flavour` before removing it in v1?

Yes but sadly enough only on reload (maybe if we get hacky with sleep and & also on first load), was thinking making a pinned issue with the breaking changes and the matching migrations that is updated for each version.

sgoudham:
A migration guide document like `0.x -> 1.x` might be nice. I'm not a fan of adding a `sleep` command lol. Even if it's just on reload, it gives people a notification right?

PR #279, opened 2024-08-06
feat: add basic install instructions

kjnsn:


vdbe:
I would use `thm_` as prefix over `ctp_` seems to be a "standard" in the tmux "community" and keeps compatibility with the plugin.
 
https://github.com/search?q=thm_++path%3A*.tmux&type=Code&ref=advsearch&l=&l=&p=1

Issue #280, opened 2024-08-06
Unable to use colon as a separator


j-xella:
### Is there an existing issue outlining your problem?

- [X] I have searched the existing issues and they do not solve my problem.

### Describe your problem.

Before, the following used to work:

```bash
set -g @catppuccin_window_middle_separator ":"
```

With recent updates, it does not work any more... Any other separators work (even double colon "::"), but not ":"

### Paste your configuration.

```sh
set -g @catppuccin_window_left_separator ""
set -g @catppuccin_window_middle_separator ":"
set -g @catppuccin_window_right_separator ""
set -g @catppuccin_window_status_enable "yes"
# set -g @catppuccin_window_status_icon_enable "no"
set -g @catppuccin_icon_window_current "null"
set -g @catppuccin_icon_window_last ""
set -g @catppuccin_window_default_text "#{s/^(TFS|KB|SUMMIT-)([0-9]*)/KB-$2/:window_name}"
set -g @catppuccin_window_current_text "#{s/^(TFS|KB|SUMMIT-)([0-9]*)/KB-$2/:window_name}"
```

### What tmux version are you seeing the issue on?

tmux built from recent commit 651891c3

vdbe:
Fixed it it in my fork wont be an issue when merging batching again.
Thank you for reporting it so fast.

PR #282, opened 2024-08-06
fixup! perf: batch tmux show-options (#240)

vdbe:
Changes the line separator used by tmux_batch from ':' to the ascii unit separator

FIXES: #240


j-xella:
That seems to fix #280

vdbe:
Have applied it to my dev branch for batching but still need to fix the escaping before trying to merge it again

PR #288, opened 2024-08-12
perf: batch tmux show

vdbe:
Try 2

### DIFF
--- a/builder/window_builder.sh
+++ b/builder/window_builder.sh
@@ -1,5 +1,8 @@
 #!/bin/sh
 
+
+show_window_status=
+
 build_window_format() {
   local number="$1"
   local color="$2"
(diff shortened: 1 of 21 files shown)


Bjorn-Eric-Abr:
Hello! Thanks for a great plugin. I forked the `main` branch of the repo to make some customisations (as you do) and found something odd.
After making my changes I wanted to customise Gitmux and it only showed the Git icon and nothing else. Reforking the whole repo (not just `main` as per Github defaults) and using `#latest` on my forked branch and the customisations work, as they do with the `#latest` (tag 0.1.0) in this repo.

**TLDR;** I think this commit makes the Gitmux customisations not work. Just wanted to give a heads up, if this is known plz ignore :)

vdbe:
@Bjorn-Eric-Abr Can you create an issue with the config used, default options from `gitmux` seem to work on the main branch (don't use it myself so don't know if anything is missing)?
image

Bjorn-Eric-Abr:
Thanks for the fast reply!
Same for me, default works but customisation does not. I'll create an issue.

Issue #297, opened 2024-08-22
Gitmux custumisation stopped working


Bjorn-Eric-Abr:
### Is there an existing issue outlining your problem?

- [X] I have searched the existing issues and they do not solve my problem.

### Describe your problem.

Hello!

In reply to the discussion on https://github.com/catppuccin/tmux/pull/288
When using the main branch the Gitmux customisation no longer works.
image

Works with default configuration, both custom and default works on tag 0.1.0, eg #latest when using TPM.

Running
tmux 3.4
Macosx 14.6.1 (Darwin arm)
Wezterm xterm-256color

For more context please look at https://github.com/catppuccin/tmux/pull/288#issuecomment-2305424427. Thanks for looking in to this, and please let me know if I got anything wrong or if you need further information.

### Paste your configuration.

```sh
# $HOME/.config/tmux/tmux.conf

set -g @catppuccin_status_modules_right "gitmux"
set -g @catppuccin_gitmux_text "#(gitmux -cfg $HOME/.gitmux.conf \"#{pane_current_path}\")"

# TPM plugins go last
set -g @plugin 'tmux-plugins/tpm'
set -g @plugin 'catppuccin/tmux#main'

# Initialize TMUX plugin manager (keep this line at the very bottom of tmux.conf)
run '~/.tmux/plugins/tpm/tpm'
```

```sh
# $HOME/.gitmux.conf
tmux:
    symbols:
        branch: '⎇ '
        hashprefix: ':'
        ahead: ↑·
        behind: ↓·
        staged: '● '
        conflict: '✖ '
        modified: '✚ '
        untracked: '… '
        stashed: '⚑ '
        clean: ✔
        insertions: Σ
        deletions: Δ
    styles:
        clear: '#[fg=default]'
        state: '#[fg=red,bold]'
        branch: '#[fg=white,bold]'
        remote: '#[fg=cyan]'
        divergence: '#[fg=default]'
        staged: '#[fg=green,bold]'
        conflict: '#[fg=red,bold]'
        modified: '#[fg=red,bold]'
        untracked: '#[fg=magenta,bold]'
        stashed: '#[fg=cyan,bold]'
        clean: '#[fg=green,bold]'
        insertions: '#[fg=green]'
        deletions: '#[fg=red]'
    layout: [branch, divergence]
    options:
        branch_max_len: 0
        branch_trim: right
        ellipsis: …
        hide_clean: false
        swap_divergence: false
        divergence_space: false
```
[...]

PR #298, opened 2024-08-23
fix: escaping in options

vdbe:
FIXES: #297


vdbe:
All configs in https://github.com/vdbe/catppuccin-tmux-dev have no regression from this change.

PR #300, opened 2024-08-28
docs: add clarifying note in the README about null values

ItzYeho:
This took me way too long to realize there even is a way to "disable" (set to empty) a customization without changing the code. This PR addresses the issue and adds a note at the top of the customization section that every text-based customization may be null.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,8 +85,15 @@ This is a diagram of how the theme is split between its components.
 ## Configuration options
 
 All flavors support certain levels of customization that match our [Catppuccin
-Style Guide][style-guide]. To add these customizations, add any of the following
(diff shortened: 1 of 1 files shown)


vdbe (changes requested):
Probably best to remove these lines since it is now documented better elsewhere 
https://github.com/catppuccin/tmux/blob/8276c5a5e33dbbbae3d370db2f6129e20b402f8a/README.md?plain=1#L343-L351

ItzYeho:
Added your requested change. I do not necessarily agree with it, since in my opinion it didn't hurt being there, but I can also see the reasoning for it.

vdbe (approved):

PR #306, opened 2024-09-04
Rewrite to use native tmux configuration

kjnsn:


catppuccin_tmux.conf
```diff
@@ -0,0 +1,105 @@
...
+set -F menu-selected-style "#{E:@catppuccin_menu_selected_style}"
+set -F menu-border-style "#{E:@catppuccin_menu_border_style}"
+
+
+# Source the status line modules
+
+# panes
+pane_format=$(load_modules "pane_default_format" "$modules_custom_path" "$modules_pane_path")
+
+set -wF pane-active-border-style "#{E:@catppuccin_active_border_style}"
```
vdbe, line 73:
```suggestion
set -wF pane-active-border-style "#{E:@catppuccin_pane_active_border_style}"
```
(also -g instead of -w)

catppuccin_tmux.conf
vdbe, line 1: only `set`'s with `-g` seem to have an effect, every other set does nothing

catppuccin_tmux.conf
```diff
@@ -0,0 +1,105 @@
...
+# menu
+set -F menu-style "#{E:@catppuccin_menu_style}"
+set -F menu-selected-style "#{E:@catppuccin_menu_selected_style}"
+set -F menu-border-style "#{E:@catppuccin_menu_border_style}"
+
+
+# Source the status line modules
+
+# panes
+pane_format=$(load_modules "pane_default_format" "$modules_custom_path" "$modules_pane_path")
```
vdbe, line 71: Had to comment out but is to be expected.

tests/harness.sh
```diff
@@ -0,0 +1,121 @@
...
+}
+
+parse_params "$@"
+setup_colors
+
+run_test() {
+  msg "Running test ${test_script} and checking the output against ${expected_output}\n"
+  msg "Using socket ${SOCKET_NAME}"
+
+  kill_tmux_server
```
vdbe, line 104: If `kill_tmux_server` fails because there is no server running it will trigger the trap which will call `kill_tmux_server` again.

image

also prints an error msg while it's expected to fail.

tmux.tera
```diff
@@ -6,20 +6,21 @@ whiskers:
...
-thm_pink="#{{ palette.pink.hex | lower }}"
-thm_red="#{{ palette.red.hex | lower }}"
-thm_green="#{{ palette.green.hex | lower }}"
-thm_yellow="#{{ palette.yellow.hex | lower }}"
-thm_blue="#{{ palette.blue.hex | lower }}"
-thm_orange="#{{ palette.peach.hex | lower }}"
-thm_black4="#{{ palette.surface2.hex | lower }}"
+set -g @thm_bg "#{{ palette.base.hex | lower }}"
+set -g @thm_fg "#{{ palette.text.hex | lower }}"
+set -g @thm_cyan "#{{ palette.sky.hex| lower }}"
```
vdbe, line 16:
```suggestion
set -g @thm_cyan "#{{ palette.sky.hex | lower }}"
```

tests/harness.sh
```diff
@@ -0,0 +1,121 @@
+#!/usr/bin/env bash
+
+set -Euo pipefail
+trap cleanup SIGINT SIGTERM ERR EXIT
+
+script_dir=$(cd "$(dirname "${BASH_SOURCE[0]}")" &>/dev/null && pwd -P)
+
+usage() {
```
vdbe, line 8:
```suggestion
usage() {
  trap - EXIT
```

No need to call cleanup on exit here since it's expected

run_tests.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+set -Eeuo pipefail
+
+script_dir=$(cd "$(dirname "${BASH_SOURCE[0]}")" &>/dev/null && pwd -P)
+
+${script_dir}/tests/harness.sh --test ${script_dir}/tests/default_options.conf --expected ${script_dir}/tests/default_options_expected.txt
```
vdbe, line 7:
```suggestion
${script_dir}/tests/harness.sh --test ${script_dir}/tests/default_options.conf --expected ${script_dir}/tests/default_options_expected.txt "$@"
```

Allows `--no-color`, overwriting expected output, etc
I guess the only practical use case for this is `--no-color` when adding more tests, which is already possible via env var.

vdbe (changes requested):
Seems to be something weird with none global options.

Did not review status/application.conf since I don't really know how to use it.

catppuccin_tmux.conf
```diff
@@ -0,0 +1,105 @@
+source -F "#{d:current_file}/themes/catppuccin_#{@catppuccin_flavor}.tmuxtheme"
+
+set -gF status-justify "#{@catppuccin_status_justify}"
```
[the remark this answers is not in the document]
kjnsn: I'm not sure either, most of this was a copy-paste from the existing config. Happy to drop this or wrap it with an %if. What do you think?

run_tests.sh
kjnsn: Ideally `--no-color` would be determined automatically (I've confirmed this works as intended with github actions and my terminal). Also not a bad idea to let this be overridden if required (or any other options added in the future)

tmux.tera
kjnsn: Hahahah I remember this

catppuccin_tmux.conf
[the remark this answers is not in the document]
vdbe: I Would drop but maybe best to port/get everything working before changing how it works since it already here.

vdbe (approved):

PR #307, opened 2024-09-04
Update README.md mistake on selecting flavour

af-lozano:
The docs say that the flavour is picked with @catppuccin_flavor, but it's really done with @catppuccin_flavour

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ set -g @plugin 'tmux-plugins/tpm'
 3. (Optional) Set your preferred flavor, it defaults to `"mocha"`:
 
 ```bash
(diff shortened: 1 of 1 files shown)


vdbe:
The spelling was changed to match other catppuccin ports. flavour still works for now (with warning) but flavor is recommended.

I think you need to update the plugin.

https://github.com/catppuccin/tmux/blob/c26d7b8ce50c1f0d9d6c41877ebf12ff8062746e/catppuccin.tmux#L48-L59

af-lozano:
@vdbe Oh sorry! thought i had the latest version due to having installed with tpm and the latest flag. But yeah, updating it ended up working. Thanks a lot.

vdbe:
No problem happy to help

PR #310, opened 2024-09-13
feat(status): Add support for status module middle separators

sebaraj:


README.md
```diff
@@ -255,6 +255,12 @@ tmux kill-server & tmux
 set -g @catppuccin_status_left_separator ""
 ```
 
+#### Set the status module middle separator:
+
+```sh
+set -g @catppuccin_status_middle_separator ""
```
vdbe, line 261:
```suggestion
set -g @catppuccin_status_middle_separator "null"
```
Setting it to `""` causes it to fallback to the default (which is in this case happens to be `""`).

vdbe (changes requested):

README.md
sebaraj: Got it, thank you for the catch.

vdbe (approved):

PR #313, opened 2024-09-15
chore: replace `@message_background` with an hidden variable

vdbe:
This does not add extra option when using `tmux show -g` (or unsetting it later) and does not require expansion with `#{@`.


kjnsn:
Can you please add a description to the PR and/or commit message explaining the "why" of this change?

PR #314, opened 2024-09-15
chore: improve testing

vdbe:
- fail tests if test scripts exit's with none 0 code
- don't expand option by default
- check all if colors are expanded
- check if module text is not expanded

### DIFF
--- a/tests/application_module.sh
+++ b/tests/application_module.sh
@@ -8,4 +8,12 @@ source "${script_dir}/helpers.sh"
 tmux source "${script_dir}/../catppuccin_options_tmux.conf"
 tmux source "${script_dir}/../catppuccin_tmux.conf"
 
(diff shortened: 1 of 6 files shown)


tests/application_module.sh
```diff
@@ -8,4 +8,12 @@ source "${script_dir}/helpers.sh"
...
 tmux source "${script_dir}/../catppuccin_tmux.conf"
 
-print_option @catppuccin_status_application | sed -E 's/\b(bash|fish|zsh)\b/<application>/'
+application_text=$(get_option @catppuccin_application_text)
+
+print_option @catppuccin_status_application | grep -q "$application_text" ||
+  echo "@catppuccin_status_application expanded @catppuccin_application_text more than once"
+
+print_option @catppuccin_status_application | grep -q "@thm_" &&
+  echo "@catppuccin_status_application did not expand all colors"
```
vdbe, line 17: maybe add a `die` function in helpers that takes a msg and exit's with 1?

PR #315, opened 2024-09-15
Feat/tmux native/status modules

vdbe:
Adds an module template that can be sourced instead of copied inside every module.

testing depends on #314


catppuccin_tmux.conf
```diff
@@ -1,5 +1,10 @@
 source -F "#{d:current_file}/themes/catppuccin_#{@catppuccin_flavor}_tmux.conf"
-source -F "#{d:current_file}/status/*.conf"
+
+source -F "#{d:current_file}/status/application.conf"
+source -F "#{d:current_file}/status/battery.conf"
+source -F "#{d:current_file}/status/clima.conf"
+source -F "#{d:current_file}/status/cpu.conf"
+source -F "#{d:current_file}/status/load.conf"
```
vdbe, line 7: Cannot use `source -F "#{d:current_file}/status/*.conf"` since all conf files in `./status` would be sourced before the source inside for example `application.conf` resulting in all sources of `utils/status_module.conf` having the (hidden) values from the last alphabetical in `./status`.

PR #318, opened 2024-09-24
Rewrite to use Tmux native configuration rather than bash

kjnsn:
TODO before merging:

- [x] Release version 0.3.0

To test checkout the tmux-native branch. You can specify `catppuccin/tmux#tmux-native` as a TPM plugin string.

So instead of `set -g @plugin "catppuccin/tmux"` replace it with `set -g @plugin "catppuccin/tmux#tmux-latest"`. 

Rewrite the plugin to use tmux native configuration files rather than bash. This will help with the speed of the plugin, allow users to use theme colours to customise their own setups, and fixes some quirks with the bash version.


catppuccin_tmux.conf
```diff
@@ -0,0 +1,279 @@
...
+source -F "#{d:current_file}/status/application.conf"
+source -F "#{d:current_file}/status/battery.conf"
+source -F "#{d:current_file}/status/clima.conf"
+source -F "#{d:current_file}/status/cpu.conf"
+source -F "#{d:current_file}/status/date_time.conf"
+source -F "#{d:current_file}/status/directory.conf"
+source -F "#{d:current_file}/status/host.conf"
+source -F "#{d:current_file}/status/load.conf"
+source -F "#{d:current_file}/status/pomodoro_plus.conf"
+source -F "#{d:current_file}/status/session.conf"
```
vdbe, line 12: Maybe better to not load all modules by default to reduce load time.

My suggestion would be to only load modules that require _no_ extra plugin since most people will only use those.
For the other modules something like `set -gF @catppuccin_dir "#{d:current_file}"`
and in the use config (`tmux.conf`) `source -F "#{@catppuccin_dir}/status/<module>.conf"`

This would also limit the effect of adding more modules and allow easy sourcing of `status_module.conf` for custom modules.

catppuccin_tmux.conf
```diff
@@ -0,0 +1,279 @@
...
+
+set status-left-length "100"
+set status-right-length "100"
+
+%if "#{==:#{@catppuccin_status_background},default}"
+  set -gF status-style "bg=#{@thm_bg},fg=#{@thm_fg}"
+  %hidden CTP_MESSAGE_BACKGROUND="#{@thm_surface_0}"
```
vdbe, line 19: Starting to see the point behind the option route since you can unset it.

utils/status_module.conf
```diff
@@ -0,0 +1,29 @@
...
+# @catppuccin_status_connect_separator != "yes"
+set -gqF @_ctp_connect "#{?#{==:#{@catppuccin_status_connect_separator},yes},, }"
+
+# Add the left separator. Background is none (switched off).
+set -gF "@catppuccin_status_${MODULE_NAME}" \
+  "#[fg=#{@catppuccin_${MODULE_NAME}_color},bg=default,nobold,nounderscore,noitalics]#{@_ctp_connect}#{@catppuccin_status_left_separator}"
+
+# Add the icon. Background is the color, and foreground is crust.
+set -agF "@catppuccin_status_${MODULE_NAME}" \
+    "#[fg=#{@thm_crust},bg=#{@catppuccin_${MODULE_NAME}_color}]#{@catppuccin_${MODULE_NAME}_icon} "
```
vdbe, line 13: Space is not removed if icon is empty, see #265.

catppuccin_tmux.conf
kjnsn: I always prefer to measure before optimising. "premature optimisation is the root of all evil". 

What's the difference in speed? Does it matter? Have you measured it?

Also for custom modules I was thinking of just letting people write their own formats, since it's so easy now.

catppuccin.tmux
```diff
@@ -3,219 +3,5 @@
...
-
-  # modes
-  setw clock-mode-colour "${thm_blue}"
-  setw mode-style "fg=${thm_pink} bg=${thm_black4} bold"
-
-  tmux "${tmux_commands[@]}"
-}
-
-main "$@"
+tmux source "${PLUGIN_DIR}/catppuccin_options_tmux.conf"
```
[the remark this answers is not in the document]
kjnsn: Hmm maybe I'm missing some context. Are you able to elaborate?

README.md
```diff
@@ -49,34 +49,42 @@
...
+1. Clone this repository to your desired location (e.g.
+   `~/.config/tmux/plugins/catppuccin`).
+
+```bash
+mkdir -p ~/.config/tmux/plugins/catppuccin
+git clone https://github.com/catppuccin/tmux.git ~/.config/tmux/plugins/catppuccin/tmux
+```
+
+2. Add the following line to your `tmux.conf` file:
+    `run ~/.config/tmux/plugins/catppuccin/tmux/catppuccin.tmux`
```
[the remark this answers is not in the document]
kjnsn: The options _cannot_ be sourced inside the main config. This is because the main config and everything it sources uses `%if` statements, which are resolved at evaluation time. Since those `%if` statements require the variables to already be set, they will not resolve correctly, hence the options need to be calculated completely _before_ the main config is evaluated.

catppuccin_tmux.conf
vdbe: - no modules: 
- basic modules: application, date_time, directory, host, session, uptime, user
- all modules/current: added gitmux, uptime, user and weather on top of what is currently sourced (were missing

image

The cost per module seems to be +1.3 ms (for me) almost negligible but it's more about putting a limit on it so it does not grow out of hand.
image

README.md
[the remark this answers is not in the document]
vdbe: Maybe it's worth adding a `catppuccin.conf`
image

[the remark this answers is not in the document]
kjnsn: Honestly, I don't think that the added complexity justifies a 20ms bump from 40ms. We could load the config as is and maintain a frame rate of 25fps, which is fast enough for people not to notice. Also it's a terminal, and this command is run once at tmux startup. I'm not going to spend time worrying about this.

catppuccin_tmux.conf
kjnsn: I get putting a limit on it, that's reasonable. I think we discussed that 200ms was a reasonable upper limit? WDYT?

vdbe: We landed on 250ms we are _well_ under that (and don't think we will ever reach it).
I just think if we draw a line on what modules to auto import plugin-less modules is the only one you can really draw.

Thinking about it how would plugins work with tpm?
They need (in most cases) `#{@catppuccin_status_<module>}` to be expanded in `status-left/right` before running but you can only expand after running the catppuccin plugin and tpm runs all plugins in sequence on `run '~/.tmux/plugins/tpm/tpm'`. Am I missing something?

Only solution I can see is in `catppuccin_tmux.conf` detect if `@plugin` is set and expand the necessary options. (or just always expand them).

README.md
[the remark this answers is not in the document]
vdbe: No problem will probably add it myself at some time and update the manual method in the readme since it is in my eyes complexity free.

catppuccin_options_tmux.conf
```diff
@@ -0,0 +1,47 @@
...
+set -ogq @catppuccin_window_status_style "basic" # basic, rounded, slanted, custom, or none
+set -ogq @catppuccin_window_default_fill "number"
+set -ogq @catppuccin_window_default_color "#{@thm_bg}"
+set -ogq @catppuccin_window_default_background "#{@thm_overlay_2}"
+set -ogq @catppuccin_window_default_text " ##T"
+set -ogq @catppuccin_window_current_fill "number"
+set -ogq @catppuccin_window_current_color "#{@thm_surface_1}"
+set -ogq @catppuccin_window_current_background "#{@thm_peach}"
+set -ogq @catppuccin_window_current_text " ##T"
+set -ogq @catppuccin_window_number_position "left"
```
vdbe, line 39: Maybe change `@catppuccin_window_current_text`/`@catppuccin_window_text` to `#W`/`#{window_name}` it's the default for tmux and there is almost every other week an issue/comment about it.

utils/status_module.conf
kjnsn: Has this been fixed? Can't remember

catppuccin_options_tmux.conf
kjnsn: This is similar to the default, but it means applications can set the window title themselves. neovim in particular utilises this, as do fish and oh-my-zsh. I agree that the current directory is a poor default. I'm happy to change this to `#W` if desired, but `#T` has been an amazing default for me

PR #319, opened 2024-09-26
feat(status): Add support for ram module

jle-quel:


kjnsn:
This looks awesome! Are you able to target the `tmux-native` branch? We're close to merging it in and it's a rewrite of the entire plugin.

jle-quel:
TODO:
- [ ] Write test
- [ ] Update README

kjnsn:
Hey! What plugin is this using? I'd probably be keen to use something like this, which would work better with (modern) tmux format strings: https://github.com/thewtex/tmux-mem-cpu-load

moetayuko:
hello, any progress with this pr? it's been several months since the last activity

> Hey! What plugin is this using?

I think it uses tmux-cpu, whose CPU percentage feature is already supported by https://github.com/catppuccin/tmux/blob/main/status/cpu.conf

> I'd probably be keen to use something like this, which would work better with (modern) tmux format strings: https://github.com/thewtex/tmux-mem-cpu-load

IMO both can be supported. tmux-mem-cpu-load requires to build a native program, while tmux-cpu is pure shell and thus more portable

kjnsn:
> hello, any progress with this pr? it's been several months since the last activity
> 

It's really trivial to add a section to the status line that shows memory usage in some way. Like, a single not very long line. You can add some examples somewhere if you'd like

moetayuko:
> > hello, any progress with this pr? it's been several months since the last activity
> 
> It's really trivial to add a section to the status line that shows memory usage in some way. Like, a single not very long line. You can add some examples somewhere if you'd like

Sure, and that's what I'm living with currently. An upstream solution would further simplify my setup.

kjnsn:
> Sure, and that's what I'm living with currently. An upstream solution would further simplify my setup.

We don't really have the capacity, as a team of one, to maintain both colours and system information shell scripts. If you want something plug and play id suggest tmux powerline

PR #320, opened 2024-09-26
feat(status/window): icons

vdbe:
- Adds support for window flags and flag icons
- Moves the spaced from before/after the icons to inside the icon default to save on boiler plate compares in tmux script.
- renames `catppuccin_icon_<icon_name>` to `catppuccin_window_flags_<icon_name>`, is this name to long?

### DIFF
--- a/catppuccin_options_tmux.conf
+++ b/catppuccin_options_tmux.conf
@@ -38,14 +38,14 @@ set -ogq @catppuccin_window_current_background "#{@thm_peach}"
 set -ogq @catppuccin_window_current_text " ##T"
 set -ogq @catppuccin_window_number_position "left"
 set -ogq @catppuccin_window_status "icon"
-set -ogq @catppuccin_window_flags_icon_last " 󰖰"
-set -ogq @catppuccin_window_flags_icon_current " 󰖯"
-set -ogq @catppuccin_window_flags_icon_zoom " 󰁌"
-set -ogq @catppuccin_window_flags_icon_mark " 󰃀"
-set -ogq @catppuccin_window_flags_icon_silent " 󰂛"
-set -ogq @catppuccin_window_flags_icon_activity " 󱅫"
-set -ogq @catppuccin_window_flags_icon_bell " 󰂞"
-# icon order: #!~[*-]MZ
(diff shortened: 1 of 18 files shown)


status/weather.conf
```diff
@@ -3,7 +3,7 @@
 
 # Requires https://github.com/xamut/tmux-weather.
 
-set -ogq @catppuccin_weather_icon ""
+set -ogq @catppuccin_weather_icon " "
```
kjnsn, line 6: nit: These icons aren't showing for me in chrome when reviewing changes. Maybe it's worth changing the icons to use hex codes (not in this change)

catppuccin_options_tmux.conf
```diff
@@ -37,6 +37,16 @@ set -ogq @catppuccin_window_current_color "#{@thm_surface_1}"
...
 set -ogq @catppuccin_window_number_position "left"
+set -ogq @catppuccin_window_status "icon"
+set -ogq @catppuccin_window_flags_icon_last " 󰖰"
+set -ogq @catppuccin_window_flags_icon_current " 󰖯"
+set -ogq @catppuccin_window_flags_icon_zoom " 󰁌"
+set -ogq @catppuccin_window_flags_icon_mark " 󰃀"
+set -ogq @catppuccin_window_flags_icon_silent " 󰂛"
+set -ogq @catppuccin_window_flags_icon_activity " 󱅫"
+set -ogq @catppuccin_window_flags_icon_bell " 󰂞"
+# icon order: #!~[*-]MZ
```
kjnsn, line 48: Can you expand on this comment a bit? It looks useful, but needs a bit more commentary

kjnsn:
Looking good!

status/weather.conf
vdbe: I would prefer to keep it as icon's so that people can directly copy/see it, hex code adds another barrier.

vdbe: I would prefer to keep it as icon's so that people can directly copy/see it, hex code adds another barrier.

catppuccin_options_tmux.conf
vdbe: Updated the comments, hope it's clear now.
TLDR that's the icon order tmux uses/displays when using text flags (`#F`)

PR #322, opened 2024-09-28
fix: pane-format

vdbe:
Options where being expanded to much

### DIFF
--- a/tests/default_options.sh
+++ b/tests/default_options.sh
@@ -8,6 +8,6 @@ source "${script_dir}/helpers.sh"
 tmux source "${script_dir}/../catppuccin_options_tmux.conf"
 tmux source "${script_dir}/../catppuccin_tmux.conf"
 
-print_option E:@catppuccin_flavor
-print_option E:@catppuccin_menu_selected_style
-print_option E:@catppuccin_pane_active_border_style
+print_option @catppuccin_flavor
+print_option @catppuccin_menu_selected_style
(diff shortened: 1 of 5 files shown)


tests/default_options.sh
```diff
@@ -8,6 +8,6 @@ source "${script_dir}/helpers.sh"
 tmux source "${script_dir}/../catppuccin_options_tmux.conf"
 tmux source "${script_dir}/../catppuccin_tmux.conf"
 
-print_option E:@catppuccin_flavor
-print_option E:@catppuccin_menu_selected_style
-print_option E:@catppuccin_pane_active_border_style
+print_option @catppuccin_flavor
+print_option @catppuccin_menu_selected_style
+print_option @catppuccin_pane_active_border_style
```
vdbe, line 13: Should check if the default options are being set correctly not how they expand.

vdbe (commented):
`set -g @catppuccin_pane_status_enabled "yes"` has currently no effect without `set -g pane-border-status top/bottom` maybe it's a good idea to to set it to top if enabled.

EDIT:
I see we have https://github.com/catppuccin/tmux/blob/d7f610c53d9d651d8c2c62b8874252de2fa9e36b/catppuccin_options_tmux.conf#L15 but it's never used.

tests/default_options.sh
kjnsn: ack

PR #336, opened 2024-10-01
Added missing pane right separator setting

khsoh:
Minor fix to replace the incorrect setup for the right side pane separator


kjnsn:
Thank you so much! Would you mind updating the tests and (optionally) adding some before & after screenshots?

khsoh:
OK. I've updated the test.

Issue #342, opened 2024-10-02
Battery status module not working


xronin01:
### Is there an existing issue outlining your problem?

- [X] I have searched the existing issues and they do not solve my problem.

### Describe your problem.

Battery status module not working

### Paste your configuration.

```sh
### Terminal
set -g default-terminal "tmux-256color"
set -ag terminal-features ",${TERM}:RGB"

### Mouse Support
set -g mouse on

### Status Bar Position
set -g status-position top

### Window Indexing
set -g base-index 1
set -g renumber-windows on

### Keymaps
unbind %
unbind '"'
bind v split-window -h
bind h split-window -v
bind r source-file $HOME/.config/tmux/tmux.conf \; display-message "Settings reloaded!"
# bind-key -n C-h if -F "#{@pane-is-vim}" 'send-keys C-h' 'select-pane -L'
# bind-key -n C-j if -F "#{@pane-is-vim}" 'send-keys C-j' 'select-pane -D'
# bind-key -n C-k if -F "#{@pane-is-vim}" 'send-keys C-k' 'select-pane -U'
# bind-key -n C-l if -F "#{@pane-is-vim}" 'send-keys C-l' 'select-pane -R'
bind-key -n C-Left if -F "#{@pane-is-vim}" 'send-keys C-Left' 'select-pane -L'
bind-key -n C-Down if -F "#{@pane-is-vim}" 'send-keys C-Down' 'select-pane -D'
bind-key -n C-Up if -F "#{@pane-is-vim}" 'send-keys C-Up' 'select-pane -U'
bind-key -n C-Right if -F "#{@pane-is-vim}" 'send-keys C-Right' 'select-pane -R'
# bind-key -n M-h if -F "#{@pane-is-vim}" 'send-keys M-h' 'resize-pane -L 3'
# bind-key -n M-j if -F "#{@pane-is-vim}" 'send-keys M-j' 'resize-pane -D 3'
# bind-key -n M-k if -F "#{@pane-is-vim}" 'send-keys M-k' 'resize-pane -U 3'
# bind-key -n M-l if -F "#{@pane-is-vim}" 'send-keys M-l' 'resize-pane -R 3'
bind-key -n M-Left if -F "#{@pane-is-vim}" 'send-keys M-Left' 'resize-pane -L 3'
bind-key -n M-Down if -F "#{@pane-is-vim}" 'send-keys M-Down' 'resize-pane -D 3'
[...]

PR #365, opened 2024-10-07
Issue #342: Checking if battery_icon literal helps

davidfmatheson:
Need the `battery_icon` format to expand as a literal for the battery plugin to find it and replace it.

### DIFF
--- a/tests/battery_module.sh
+++ b/tests/battery_module.sh
@@ -8,10 +8,4 @@ source "${script_dir}/helpers.sh"
 tmux source "${script_dir}/../catppuccin_options_tmux.conf"
 tmux source "${script_dir}/../catppuccin_tmux.conf"
 
-print_option @catppuccin_battery_icon | grep -q "#{battery_icon}" ||
-  echo "@catppuccin_status_battery expanded #{battery_icon} in @catppuccin_battery_icon"
-
-NL='
-'
-[[ "$(print_option @catppuccin_battery_icon)" == "${NL}@catppuccin_battery_icon #{l:#{battery_icon}} " ]] ||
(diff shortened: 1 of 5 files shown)


kjnsn:
Would you mind adding some tests?

davidfmatheson:
> Would you mind adding some tests?

I'll do my best! Still new to tmux config.

tests/battery_module.sh
```diff
@@ -10,3 +10,6 @@ tmux source "${script_dir}/../catppuccin_tmux.conf"
 
 print_option @catppuccin_battery_icon | grep -q "#{battery_icon}" ||
   echo "@catppuccin_status_battery expanded #{battery_icon} in @catppuccin_battery_icon"
+
+print_option @catppuccin_battery_icon | grep -q "#{l:#{battery_icon}}" ||
```
kjnsn, line 14: Can you add an expected output and match against that instead? I don't trust grep as much, it's not clear what is actually being printed by the option

davidfmatheson: Added an expected output match, had to resort to defining a newline char for expansion in the match because `print_option` adds a newline.

kjnsn: I'm not seeing a change to `tests/battery_module_expected.txt`

kjnsn: Also for reference I'm pretty sure `\n` works

davidfmatheson: Ah, I misunderstood the ask. I'll go take a look at the expected file.

davidfmatheson: Is this how you use the `_expected.txt` file?

kjnsn:
Thanks so much!

Issue #345, opened 2024-10-03
Weather module not working


jeffcap1:
### Is there an existing issue outlining your problem?

- [X] I have searched the existing issues and they do not solve my problem.

### Describe your problem.

The weather module does not appear to be working after upgrading to v1.0.0. On v0.3.0 I was using `tmux-weather`. After I upgraded to v1.0.0 I noticed the weather module stopped working and thought it might have been an issue with `tmux-weather` so I switched to `tmux-clima`, however, I'm seeing the same issue with that module as well. 

### Paste your configuration.

```sh
# new leader key
unbind C-b
set -g prefix C-Space
bind C-Space send-prefix

# fix colors
set -g default-terminal 'tmux-256color'
set -ga terminal-overrides ",xterm-256color*:Tc"

# cursor
# Force cursor shape as blinking bar (xterm)
# https://neovim.io/doc/user/term.html#tui-cursor-shape
set -as terminal-overrides '*:Ss=\E[%p1%d q:Se=\E[5 q'

# refresh tmux with leader + r
unbind r
bind r command-prompt "rename-window %%"
bind R source ~/.config/tmux/tmux.conf

# splits and navigation
bind '"' split-window -v -c "#{pane_current_path}"
bind % split-window -h -c "#{pane_current_path}"
bind c new-window -c "#{pane_current_path}"
bind -n M-H previous-window
bind -n M-L next-window
bind-key x kill-pane # skip "kill-pane 1? (y/n)" prompt

# Start windows and panes at 1, not 0
set -g base-index 1                  # start indexing windows at 1 instead of 0
set -g pane-base-index 1             # start indexing panes at 1 instead of 0
set-window-option -g pane-base-index 1
set -g renumber-windows on           # renumber all windows when any window is closed

# options
set -g mouse on                      # mouse support
set -g detach-on-destroy off         # don't exit from tmux when closing a session
set -g escape-time 0                 # zero-out escape time delay
set -g history-limit 1000000         # increase history size (from 2,000)
set -g set-clipboard on              # use system clipboard

# status bar
set -g status-justify left
set -g status-left-length 200
set -g status-right-length 200
set -g status-position top

# set vi mode
set -g xterm-keys on
set -g mode-keys vi
set -g status-keys vi

# act like vim
bind-key -n M-h select-pane -L
bind-key -n M-j select-pane -D
bind-key -n M-k select-pane -U
bind-key -n M-l select-pane -R

# copy mode
[...]

PR #366, opened 2024-10-08
docs: Issue #345 - clarifying steps for status modules

jeffcap1:
## Description

Adds some additional clarity for what is required when setting the status modules.

> Based on the discussion in Issue #345.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,9 +81,6 @@ git clone https://github.com/catppuccin/tmux.git ~/.config/tmux/plugins/catppucc
 <!-- x-release-please-start-version -->
 
 ```bash
-# Load catppuccin theme
-run ~/.config/tmux/plugins/tmux/catppuccin.tmux
-# ... catppuccin settings
 set -g @plugin 'catppuccin/tmux#v1.0.1' # See https://github.com/catppuccin/tmux/tags for additional tags
 # ...alongside
 set -g @plugin 'tmux-plugins/tpm'
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ git clone https://github.com/catppuccin/tmux.git ~/.config/tmux/plugins/catppucc
    listed in [Configuration Options](#configuration-options). These options must be added above the `run ~/.config...` line.
 4. Reload Tmux by either restarting or reloading with `tmux source ~/.tmux.conf`
 
-> [!IMPORTANT]  
+> [!IMPORTANT]
 > You may have to run `~/.config/tmux/plugins/tpm/bin/clean_plugins` if upgrading from an earlier version
 > (especially from `v0.3.0`).
 
@@ -81,6 +81,9 @@ git clone https://github.com/catppuccin/tmux.git ~/.config/tmux/plugins/catppucc
 <!-- x-release-please-start-version -->
 
 ```bash
+# Load catppuccin theme
+run ~/.config/tmux/plugins/tmux/catppuccin.tmux
+# ... catppuccin settings
 set -g @plugin 'catppuccin/tmux#v1.0.1' # See https://github.com/catppuccin/tmux/tags for additional tags
 # ...alongside
 set -g @plugin 'tmux-plugins/tpm'
@@ -148,6 +151,7 @@ set -g @catppuccin_window_current_background "#{@thm_mauve}"
 
 # Load catppuccin
 run ~/.config/tmux/plugins/catppuccin/tmux/catppuccin.tmux
+# For TPM, instead use `run ~/.config/tmux/plugins/tmux/catppuccin.tmux`
 
 # Make the status line pretty and add some modules
 set -g status-left ""
--- a/status/README.md
+++ b/status/README.md
@@ -1,4 +1,3 @@
-
 ### Customizing modules
 
 Every module (except the module "session") supports the following overrides:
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -81,6 +81,9 @@ git clone https://github.com/catppuccin/tmux.git ~/.config/tmux/plugins/catppucc
 <!-- x-release-please-start-version -->
 
 ```bash
+# Load catppuccin theme
+run ~/.config/tmux/plugins/tmux/catppuccin.tmux
```
kjnsn, line 85: Couple of things:

1. This example config doesn't use any status line modules that require TPM to be run twice
2. The `run` lines are in the wrong spots

status/README.md
```diff
@@ -34,6 +33,14 @@ Ex:
 set -g @catppuccin_date_time_icon "null"
 ```
 
+> [!IMPORTANT]
```
kjnsn, line 36: Why is this inside "Removing a specific module option"?

status/README.md
```diff
@@ -34,6 +33,14 @@ Ex:
...
+> Make sure to load the catppuccin theme prior to setting your status modules.<br/>
```
kjnsn, line 37: Should be "prior to setting status-left and/or status-right"

status/README.md
```diff
@@ -34,6 +33,14 @@ Ex:
...
+> Use `run ~/.config/tmux/plugins/catppuccin/tmux/catppuccin.tmux`.<br/>
+> For TPM, instead use `run ~/.config/tmux/plugins/tmux/catppuccin.tmux`.
```
kjnsn, line 39: Should this be an instruction to run tpm?

status/README.md
```diff
@@ -34,6 +33,14 @@ Ex:
...
+
+> [!NOTE]
+> For TPM users, execute the TPM start command `run '~/.tmux/plugins/tpm/tpm'` after setting your status modules.
```
kjnsn, line 42: This is already covered in the IMPORTANT section above

status/README.md
kjnsn, line 37: Can you add an explanation as well?

kjnsn (changes requested):

status/README.md
jeffcap1: After the issue was closed I did some testing in my config. I found running just the catppuccin command first and then running the TPM command later worked as expected. That made more sense to me than running TPM twice, but I can put that in the docs if that is what you prefer.

README.md
```diff
@@ -148,6 +148,7 @@ set -g @catppuccin_window_current_background "#{@thm_mauve}"
 
 # Load catppuccin
 run ~/.config/tmux/plugins/catppuccin/tmux/catppuccin.tmux
+# For TPM, instead use `run ~/.config/tmux/plugins/tmux/catppuccin.tmux`
```
kjnsn, line 151: This is the exact same line as above, right?

status/README.md
```diff
@@ -34,6 +33,29 @@ Ex:
 set -g @catppuccin_date_time_icon "null"
 ```
 
+### Notes for TPM users
+
+Make sure you load the catppuccin theme prior to setting the status-left and/or status-left options. This allows the catppuccin settings to be defined in the global space so they can then be set. After these settings have been configured make sure to run the TPM command to load the modules. This runs the plugin and sets them into the status line.
```
kjnsn, line 38: Suggestion:

Make sure you load the catppuccin theme prior to setting the status-left and/or status-left options. This ensures the catppuccin options (such as colors and status modules) are defined  so they can then be used. 

After status-left/status-right have been set, make sure to run TPM to load the modules. This runs any plugins that may replace text in the status line.

status/README.md
```diff
@@ -34,6 +33,29 @@ Ex:
...
+
+```bash
+# ... load catppuccin theme
+run '~/.config/tmux/plugins/tmux/catppuccin.tmux' # or where this file is located on your machine
+
+# ... catppuccin settings
```
kjnsn, line 44: The `...` at the beginning isn't really needed, see https://www.thepunctuationguide.com/ellipses.html.

`...` is great when it's continuing on from a previous line and thought, for example:

```
# Load the catppuccin theme...
run '~/.config/tmux/plugins/tmux/catppuccin.tmux'

# ... and then set status-left & status-right
set -g status-left
```

I do apologise if this is overly pedantic! I'm keen to get this in a good place to help others :)

kjnsn (changes requested):

README.md
jeffcap1: You are correct, they are the same script, but the location differs. I was attempting to call out because if you are scanning the docs it's easy to miss. Especially, if you don't realize that TPM places the theme in a `tmux` directory instead of `catppuccin` and don't read the manual installation step earlier in the docs.

status/README.md
jeffcap1: That reads much better, I'll update with that suggestion.

status/README.md
jeffcap1: No worries! I thought that was a stylistic choice in the comments on the main README and was trying to keep it consistent. I'll make that adjustment.

khsoh:
Hi, I found that there are issues with the status bar when scripts are required to be executed (like the cpu status module).  I am not sure if this is a TMUX or TPM or catppuccin-tmux issue.  For example, if I use the CPU status module, after the initial start up with `tmux new-session` , the CPU status does not appear.  

I have to  reload the config with `tmux source ~/.config/tmux/tmux.conf` to make the CPU status appear.

The workaround I applied was to run:

`tmux new \; set-hook -g session-created "source ~/.config/tmux/tmux.conf"
`

The suggested method to run catppuccin.tmux first did not work for me.

kjnsn:
> Hi, I found that there are issues with the status bar when scripts are required to be executed (like the cpu status module). I am not sure if this is a TMUX or TPM or catppuccin-tmux issue. For example, if I use the CPU status module, after the initial start up with `tmux new-session` , the CPU status does not appear.

Hey! Not sure how this is relevant to this review? There are issues already filed regarding this. If you have a suggestion on how to improve the docs, please make a suggestion directly

README.md
kjnsn: Oh wow, you have no idea how long I stared at this. Thanks!

kjnsn (approved):

PR #367, opened 2024-10-08
style(status): use @catppuccin_status_background as background color

dreknix:
When changing the background of the status bar, I think the default background color of all modules should be this color. 
 
Therefore, the hard coded references to `default` and `thm_surface_1` in `utils/status_module.conf` should be in my opinion changed to `${CTP_MESSAGE_BACKGROUND}`. With this changes the background color of the status bar is always the same. 
 
I'm not so familiar with the new tmux configuration style. I hope that this is the correct way to solve the problem.

### DIFF
--- a/catppuccin_options_tmux.conf
+++ b/catppuccin_options_tmux.conf
@@ -52,4 +52,4 @@ set -ogq @catppuccin_status_middle_separator ""
 set -ogq @catppuccin_status_right_separator "█"
 set -ogq @catppuccin_status_connect_separator "yes" # yes, no
 set -ogq @catppuccin_status_fill "icon"
-set -ogq @catppuccin_status_default_background "#{E:@thm_surface_1}"
+set -ogq @catppuccin_status_default_background "#{@thm_surface_1}"
--- a/catppuccin_options_tmux.conf
+++ b/catppuccin_options_tmux.conf
@@ -52,3 +52,4 @@ set -ogq @catppuccin_status_middle_separator ""
 set -ogq @catppuccin_status_right_separator "█"
 set -ogq @catppuccin_status_connect_separator "yes" # yes, no
 set -ogq @catppuccin_status_fill "icon"
+set -ogq @catppuccin_status_default_background "#{@thm_surface_1}"
--- a/catppuccin_options_tmux.conf
+++ b/catppuccin_options_tmux.conf
@@ -52,4 +52,3 @@ set -ogq @catppuccin_status_middle_separator ""
 set -ogq @catppuccin_status_right_separator "█"
 set -ogq @catppuccin_status_connect_separator "yes" # yes, no
 set -ogq @catppuccin_status_fill "icon"
-set -ogq @catppuccin_status_default_background "#{@thm_surface_1}"
--- a/catppuccin_options_tmux.conf
+++ b/catppuccin_options_tmux.conf
@@ -52,4 +52,4 @@ set -ogq @catppuccin_status_middle_separator ""
 set -ogq @catppuccin_status_right_separator "█"
 set -ogq @catppuccin_status_connect_separator "yes" # yes, no
 set -ogq @catppuccin_status_fill "icon"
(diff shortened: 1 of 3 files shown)


kjnsn (commented):
Thanks for taking the time to make a PR!

Are you fixing a specific bug? And/or do you have before and after screenshots? 

Why is default wrong? I agree that hardcoding surface 1 should probably change.

Also where did `${CTP_MESSAGE_BACKGROUND}` come from? Where is that environment variable set?

dreknix:
I want the status bar to have a consistent color, when setting `@catpuccin_status_background`, e.g., to `@thm_surface_0`. The status bar and all modules should have the same background color.

Before

before

After

after

The variable `${CTP_MESSAGE_BACKGROUND}` comes from the file `catppuccin_tmux.conf`. The content of the variable depends on the setting of `@catppuccin_status_background`.

But maybe it is better to change the hardcoded `@thm_surface_1` into a tmux variable like `@catppuccin_status_default_background`.

kjnsn:
> I want the status bar to have a consistent color, when setting `@catpuccin_status_background`, e.g., to `@thm_surface_0`. The status bar and all modules should have the same background color.
>

Yep sounds reasonable.


> The variable `${CTP_MESSAGE_BACKGROUND}` comes from the file `catppuccin_tmux.conf`. The content of the variable depends on the setting of `@catppuccin_status_background`.

Ah that's meant to be local and not exported.

> But maybe it is better to change the hardcoded `@thm_surface_1` into a tmux variable like `@catppuccin_status_default_background`.

What happens if you use default? The status line is already styled, I don't want to duplicate that if possible

dreknix:
Here is my minimal `tmux.conf`

```
set -g default-terminal 'screen-256color'
set -g base-index 1
set -g pane-base-index 1

set-option -g status-position top

set -g @catppuccin_status_background "#{@thm_surface_0}"

run ~/.local/share/tmux/plugins/tmux/catppuccin.tmux

set -g status-left "#{E:@catppuccin_status_session}"
set -g status-right "#{E:@catppuccin_status_host}"
set -ag status-right "#{E:@catppuccin_status_date_time}"
```

dreknix:
> > The variable `${CTP_MESSAGE_BACKGROUND}` comes from the file `catppuccin_tmux.conf`. The content of the variable depends on the setting of `@catppuccin_status_background`.
> 
> Ah that's meant to be local and not exported.

Ok, good to know.

> > But maybe it is better to change the hardcoded `@thm_surface_1` into a tmux variable like `@catppuccin_status_default_background`.
> 
> What happens if you use default? The status line is already styled, I don't want to duplicate that if possible

When I use `default` than in line 27 in `utils/status_module.conf` the default foreground color is used. Therefore, a variable is needed which stores the default color of the status bar if set.

Since `${CTP_MESSAGE_BACKGROUND}` should be local, I will change my PR, to reflect this discussion.

dreknix:
I changed the PR to use a new variable `@catppuccin_status_default_background` instead of the hard coded color `@thm_surface_1`. With this change the status bar and the modules can have the same color by setting both variables:

```
set -g @catppuccin_status_background '#{@thm_surface_0}'
set -g @catppuccin_status_default_background '#{@thm_surface_0}'
```

Thanks a lot for the discussion and the great plugin.

catppuccin_options_tmux.conf
```diff
@@ -52,3 +52,4 @@ set -ogq @catppuccin_status_middle_separator ""
 set -ogq @catppuccin_status_right_separator "█"
 set -ogq @catppuccin_status_connect_separator "yes" # yes, no
 set -ogq @catppuccin_status_fill "icon"
+set -ogq @catppuccin_status_default_background "#{E:@thm_surface_1}"
```
kjnsn, line 55:
```suggestion
set -ogq @catppuccin_status_default_background "#{@thm_surface_1}"
```

kjnsn:
Thank you so much!

PR #373, opened 2024-10-14
fix(weather status): Use tmux-weather's cache var

merlinblack:
After upgrading I found my weather status was not working.

xamut/tmux-weather looks for '#{weather}' in status-right to replace with the downloaded weather data.

However it can't find it as status-right will only contain '#{E:@catppuccin_status_weather}', and results in only the text '#{weather}' being displayed.

So instead of using the literal '#{weather}' in the status module, use the var that xamut/tmux-wether uses to cache it's results.

Tested on my own systems.

### DIFF
--- a/status/weather.conf
+++ b/status/weather.conf
@@ -5,6 +5,6 @@
 
 set -ogq @catppuccin_weather_icon " "
 set -ogqF @catppuccin_weather_color "#{@thm_yellow}"
-set -ogq @catppuccin_weather_text "#{l:#{weather}}"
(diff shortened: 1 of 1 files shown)


merlinblack:
Agrh - this does not quite work. Working on a better fix.

merlinblack:
Turns out the easiest way to get this going to is to add '#{E:@catppuccin_status_weather}' to the status-right using the -F flag so that #{weather} ends up in thier, and tmux-weather can then update that to call the actual weather script.

Maybe just needs a doc update.

PR #374, opened 2024-10-14
docs(README): fix incorrect link for Pocco

IroncladDev:
Looks like the Pocco81 link at the bottom links to the wrong github profile.


rogeruiz (approved):
Good catch.

PR #375, opened 2024-10-15
feat: add tmux-mem-cpu-load status

mdetweil-amzn:
- Add in [`tmux-mem-cpu-load`](https://github.com/thewtex/tmux-mem-cpu-load) status
- update tests to run a simple loop
- Add documentation.
image


kjnsn:
Oh this is awesome!!

My only hesitation is that we already have status line segments for CPU usage and for load, so this is duplicating that information. Is is possible to just output the memory usage? I also see that `tmux-mem-cpu-load` has support for displaying a gradient of colors across the usage bar, it would be great if this could somehow be mapped to the catppuccin red -> lavender spectrum.

mdetweil-amzn:
> Oh this is awesome!!
> 
> My only hesitation is that we already have status line segments for CPU usage and for load, so this is duplicating that information. Is is possible to just output the memory usage? I also see that `tmux-mem-cpu-load` has support for displaying a gradient of colors across the usage bar, it would be great if this could somehow be mapped to the catppuccin red -> lavender spectrum.

Yes, you do support CPU usage/load. There are two other PRs open about memory, both more or less pointing to this as an implementation option, but they seemed abandoned, and as I already used the tool before porting over to `catppuccin/tmux` it was an easy enough implementation. Now, I find that what presents the best experience for users is a host of options. I think each of the cpu/mem tools output slightly differently, and that can give a really custom experience. 

Maybe what that means is that the structure should start to separate out types of status into consolidated groups for each reading/decision making. With that said it does not appear possible to only output memory usage at this time. I can follow up with a second PR to implement `tmux-mem`

The `--colors` flag turns on color support, which by default changes the color based on the load. As for the gradient support, at this time it appears thats only available to powerline users, and it only supports the standard 256 color mode. So those colors are likely out of the picture for the time being.

On a completely separate separate note, I took the time to do a minor refactor of the test runner. I hope that was okay.

PR #380, opened 2024-10-17
fix: replace @_ctp_status_bg with @thm_surface_0 in module configurtion to properly show module separation

xafarr:
**Before the change:**
With `@catppuccin_status_connect_separator` set to `yes` or `no`
Screenshot 2024-10-17 at 12 04 22 pm

**After the change:**
With `@catppuccin_status_connect_separator` set to `yes`
Screenshot 2024-10-17 at 12 09 50 pm

With `@catppuccin_status_connect_separator` set to `no`
Screenshot 2024-10-17 at 12 08 59 pm

### DIFF
--- a/utils/status_module.conf
+++ b/utils/status_module.conf
@@ -5,6 +5,12 @@
 set -gqF @_ctp_connect_style \
   "#{?#{==:#{@catppuccin_status_connect_separator},yes},,#[bg=default]}"
 
+%if "#{==:#{@catppuccin_status_${MODULE_NAME}_bg_color},}"
+  set -gqF @_ctp_module_bg_color "#{@catppuccin_status_module_bg_color}"
+%else
+  set -gqF @_ctp_module_bg_color "#{@catppuccin_status_${MODULE_NAME}_bg_color}"
+%endif
+
 set -gF "@catppuccin_status_${MODULE_NAME}" \
   "#[fg=#{@catppuccin_${MODULE_NAME}_color},nobold,nounderscore,noitalics]#{@_ctp_connect_style}#{@catppuccin_status_left_separator}"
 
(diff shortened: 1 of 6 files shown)


utils/status_module.conf
```diff
@@ -16,15 +16,15 @@ set -agF "@catppuccin_status_${MODULE_NAME}" "#[fg=#{@catppuccin_${MODULE_NAME}_
 # If _only_ the icon should be filled in, then change the background
 # to catppuccin_status_default_background, and the foreground to crust. Otherwise leave the formatting as-is.
 %if "#{==:#{@catppuccin_status_fill},icon}"
-  set -agF "@catppuccin_status_${MODULE_NAME}" "bg=#{E:@_ctp_status_bg}]#{@catppuccin_status_middle_separator}#[fg=#{@thm_fg}] "
+  set -agF "@catppuccin_status_${MODULE_NAME}" "bg=#{@thm_surface_0}]#{@catppuccin_status_middle_separator}#[fg=#{@thm_fg}] "
```
kjnsn, line 19:
```suggestion
  set -agF "@catppuccin_status_${MODULE_NAME}" "bg=#{@catppuccin_status_${MODULE_NAME}_bg_color]#{@catppuccin_status_middle_separator}#[fg=#{@thm_fg}] "
```

Funnily enough I merged a PR a couple of weeks ago that did the opposite of this, which leads me to the conclusion that the background colors of modules should be configurable.

What do you think of having a `${MODULE_NAME}_bg_color` variable which defaults to `thm_surface_0`? Otherwise there could be a global background status segment color, but that is less flexible

xafarr: In my opinion, we keep the global background status segment color for consistency among different modules. We already have the module color, and I'm not sure if we want to have another color for the background when it's an icon-fill only. But on the other hand, it's a lot more flexible for the user to decide how their status line should look like. It just adds an extra bit of maintenance for all the modules. If we are to provide that flexibility, I'd add the option of controlling it globally and enabling overriding it for modules if wanted.

Just thinking out loud. If we're providing a way to change the background color, then we would have to have a way to change the foreground color. What do you think?

 Let me know what you're leaning towards.

xafarr: Following is the code snippet I tested to implement the below idea.

> If we are to provide that flexibility, I'd add the option of controlling it globally and enabling overriding it for modules if wanted.

```tmux
%if "#{==:#{@catppuccin_status_${MODULE_NAME}_bg_color},}"
  set -gqF @_ctp_module_bg_color "#{@catppuccin_status_module_bg_color}"
%else
  set -gqF @_ctp_module_bg_color "#{@catppuccin_status_${MODULE_NAME}_bg_color}"
%endif
```
And global value in `catppuccin_options_tmux.conf`
```tmux
set -ogq @catppuccin_status_module_bg_color "#{@thm_surface_0}"
```

kjnsn: That sounds like a really good solution to me :)

xafarr: Great. I'll update the PR.

bvelastegui (commented):
Nice work!

kjnsn:
Thanks so much! This will be released in v2

PR #383, opened 2024-10-18
Fix up the docs to make configuration match screenshot

davidfmatheson:
The screenshot has blue background for numbers on non-current windows, and space between text and number on right. Update sample config to make sure it matches result in screenshot. Without `@thm_blue` it will default to grey.

### DIFF
--- a/README.md
+++ b/README.md
@@ -420,10 +420,12 @@ set -g @catppuccin_window_status_style "rounded"
 set -g @catppuccin_window_number_position "right"
 
 set -g @catppuccin_window_default_fill "number"
-set -g @catppuccin_window_default_text "#W"
+set -g @catppuccin_window_default_text "#W "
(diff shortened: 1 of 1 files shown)


kjnsn:
Yeah the screenshot needs updating. The colors were changed to improve accessibility. Would you mind updating the screenshot?

davidfmatheson:
I was updating the config to what I thought would generate the screenshot. You'd rather the screenshot be updated to match what was in the config without changes?

kjnsn:
Yes

davidfmatheson:
Ok, replaced screenshot, and removed the default peach/blue in config. I left the space after `#W` because that is necessary to make sure the text doesn't bump up against the right-hand numbers.

kjnsn:
Thank you so much!!

Issue #389, opened 2024-10-20
slanted window status style with right aligned number leading to gap


NielsCloer:
### Is there an existing issue outlining your problem?

- [X] I have searched the existing issues and they do not solve my problem.

### Describe your problem.

Problem as described in title.
I have set window number position to "right" and window status style to "slanted". This ends up leaving a gap to the left of the number.
As far as I have been able to identify this isn't what is supposed to happen.

### Paste your configuration.

```
unbind r
bind r source-file ~/.tmux.conf

set -g mouse on

bind-key h select-pane -L
bind-key j select-pane -D
bind-key k select-pane -U
bind-key l select-pane -R
bind-key x kill-pane
set -g base-index 1

set-option -g status-position top

set -g default-terminal "tmux-256color"

# Configure the catppuccin plugin
set -g @catppuccin_flavor "macchiato"

set -g @catppuccin_window_number_position "right"
set -g @catppuccin_window_status_style "slanted"

set -g @catppuccin_window_default_fill "number"
set -g @catppuccin_window_default_text "#W "

set -g @catppuccin_window_current_fill "number"
set -g @catppuccin_window_current_text "#{pane_current_path}"

set -g @catppuccin_status_left_separator  ""
set -g @catppuccin_status_right_separator " "
set -g @catppuccin_status_fill "icon"
set -g @catppuccin_status_connect_separator "no"

run ~/.config/tmux/plugins/catppuccin/tmux/catppuccin.tmux


set -g status-left ""
set -g status-right "#{E:@catppuccin_status_user}"
set -ag status-right "#{E:@catppuccin_status_directory}"

# List of plugins
set -g @plugin 'tmux-plugins/tpm'


# Other examples:
# set -g @plugin 'github_username/plugin_name'
# set -g @plugin 'github_username/plugin_name#branch'
# set -g @plugin 'git@github.com:user/plugin'
# set -g @plugin 'git@bitbucket.com:user/plugin'

# Initialize TMUX plugin manager (keep this line at the very bottom of tmux.conf)
run '~/.tmux/plugins/tpm/tpm'
```

### Attach screenshots.

![image](https://github.com/user-attachments/assets/8109fc06-5253-4561-a25e-3121ac2155ef)

### What tmux version are you seeing the issue on?

tmux 3.2a

kjnsn:
This is a known issue that will be fixed in version 2. 

Also out of curiosity, why are you using tmux if your terminal supports tabs?

[...]

PR #390, opened 2024-10-20
Issue #389: Need dynamic middle separator when slanted

davidfmatheson:
Saw this may be fixed in a v2.0.0, but I think this'll fix just issue #389, if you want it.


kjnsn:
It is fixed in version 2 (check the v2 branch), which should be merged in the next day or two. 

I don't really want to make another release before then. I guess people could copy and paste the stuff that works from that branch if they really wanted to?

PR #391, opened 2024-10-20
Fix cpu status in wrong format

unisons:
cpu_bg_color is replaced by "#[bg=green]" by the plugin. But the expecting value from catppuccin is just "green" that is just color value.


unisons:
It seems that the color is correct but there is some side effect with v2 in my setup.

PR #394, opened 2024-10-21
Fix session status icon bg color change by client_prefix

unisons:
The bg color of icon is always green.
I think the expands should be done later. There might be more cases.
This bug is observed on v2 but not v1.0.3.

### DIFF
--- a/status/session.conf
+++ b/status/session.conf
@@ -2,7 +2,7 @@
 %hidden MODULE_NAME="session"
 
 set -ogq "@catppuccin_${MODULE_NAME}_icon" " "
-set -ogqF "@catppuccin_${MODULE_NAME}_color" "##{?client_prefix,#{E:@thm_red},#{E:@thm_green}}"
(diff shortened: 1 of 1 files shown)


kjnsn:
Thanks so much!

Also why on earth do we change the color of the _session_ when the _prefix_ key is pressed, somehow I missed this.

89iuv:
> Thanks so much!
> 
> Also why on earth do we change the color of the _session_ when the _prefix_ key is pressed, somehow I missed this.

I really liked this feature. 
It is an indicator that the tmux prefix has been triggered and any other input will be part of a tmux shortcut.

kjnsn:
Fair enough, but what does it have to do with the session name?

89iuv:
> Fair enough, but what does it have to do with the session name?

It should not have anything to do with the session name. 
I think the logic brakes with some themes.

I am using it like this:
image
image

It is just a small indicator that the tmux prefix has been pressed.
It was here from the begging of the theme and i find it usefull.

PR #399, opened 2024-10-22
Update status_module.conf

davidfmatheson:
Status bar entries have text up against the icon:
image

Probably should look like:
image

### DIFF
--- a/status/directory.conf
+++ b/status/directory.conf
@@ -3,6 +3,6 @@
 
 set -ogq "@catppuccin_${MODULE_NAME}_icon" " "
 set -ogqF "@catppuccin_${MODULE_NAME}_color" "#{E:@thm_rosewater}"
-set -ogq "@catppuccin_${MODULE_NAME}_text" " bigdirectory"
+set -ogq "@catppuccin_${MODULE_NAME}_text" " #{b:pane_current_path}"
 
 source -F "#{d:current_file}/../utils/status_module.conf"
--- a/status/directory.conf
+++ b/status/directory.conf
@@ -3,6 +3,6 @@
 
 set -ogq "@catppuccin_${MODULE_NAME}_icon" " "
 set -ogqF "@catppuccin_${MODULE_NAME}_color" "#{E:@thm_rosewater}"
-set -ogq "@catppuccin_${MODULE_NAME}_text" " #{b:pane_current_path}"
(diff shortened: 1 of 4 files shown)


kjnsn:
Can you update the individual status modules? Putting a space there for all modules reduces flexibility

vdbe:
Can you share the config used?
It doesn't seem to happen on the main branch
image
```tmux
set -g default-terminal screen-256color
set -g @test_config "tmux"
set -g mouse on

run "$CATPPUCCIN_TMUX_PLUGIN"

set -g status-right-length 100
set -g status-right "#{E:@catppuccin_status_directory}"
set -ag status-right "#{E:@catppuccin_status_date_time}"
```
image

If this is a bug it could/should be fixed by using the `@catppuccin_status_middle_separator` options/setting it's default

davidfmatheson:
> Can you share the config used? It doesn't seem to happen on the main branch image
> 
> ```shell
> set -g default-terminal screen-256color
> set -g @test_config "tmux"
> set -g mouse on
> 
> run "$CATPPUCCIN_TMUX_PLUGIN"
> 
> set -g status-right-length 100
> set -g status-right "#{E:@catppuccin_status_directory}"
> set -ag status-right "#{E:@catppuccin_status_date_time}"
> ```
> 
> image
> 
> If this is a bug it could/should be fixed by using the `@catppuccin_status_middle_separator` options/setting it's default

You're right, I was setting the text with no space in my config:
```
set -g @catppuccin_directory_text "#{pane_current_path}"
set -g @catppuccin_date_time_text "%H:%M"
```

and your default config puts the space in those variables. Not sure if you want to be enable folks to set the text of those variables and still have space from the icon, but this was mostly my fault.

kjnsn:
People should still be able to set the text and have space. Does this not work?

```
set -g @catppuccin_directory_text " #{pane_current_path}"
set -g @catppuccin_date_time_text " %H:%M"
```

davidfmatheson:
> People should still be able to set the text and have space. Does this not work?
> 
> ```
> set -g @catppuccin_directory_text " #{pane_current_path}"
> set -g @catppuccin_date_time_text " %H:%M"
> ```

That does work, it's just in previous versions when I set the text without space, there would still be space between the icon and text, so I missed this in my testing.

PR #405, opened 2024-10-23
build: allow release-please to release on more scopes

sgoudham:
This PR adjusts the changelog sections so that `perf`, `revert` and `docs` commits are tracked in the CHANGELOG and are suitable candidates for the creation of a release please PR.

In an ideal world, the commits included in the CHANGELOG and the scopes responsible for the creation of the release please PR would be decoupled but it seems as though that is not currently possible.

### DIFF
--- a/.github/release-please-config.json
+++ b/.github/release-please-config.json
@@ -15,7 +15,6 @@
   },
   "changelog-sections": [
     { "type": "feat", "section": "Features" },
-    { "type": "feature", "section": "Features" },
     { "type": "fix", "section": "Bug Fixes" },
     { "type": "perf", "section": "Performance Improvements" },
     { "type": "revert", "section": "Reverts" },
--- a/.github/release-please-config.json
+++ b/.github/release-please-config.json
@@ -13,5 +13,13 @@
       ]
     }
   },
+  "changelog-sections": [
+    { "type": "feat", "section": "Features" },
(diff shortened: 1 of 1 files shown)


.github/release-please-config.json
```diff
@@ -13,5 +13,13 @@
       ]
     }
   },
+  "changelog-sections": [
+    { "type": "feat", "section": "Features" },
+    { "type": "feature", "section": "Features" },
+    { "type": "fix", "section": "Bug Fixes" },
+    { "type": "perf", "section": "Performance Improvements" },
```
sgoudham, line 20: Not sure if we need this one but I figured I'd let someone else decide.

.github/release-please-config.json
```diff
@@ -13,5 +13,13 @@
       ]
     }
   },
+  "changelog-sections": [
+    { "type": "feat", "section": "Features" },
+    { "type": "feature", "section": "Features" },
```
sgoudham, line 18: Might actually remove this since it's a bit redundant and it feels like providing an option that I don't actually want anyone to use :thinking:

.github/release-please-config.json
kjnsn: LGTM

.github/release-please-config.json
sgoudham, line 18:
```suggestion
```

PR #416, opened 2024-10-28
Extension to support user custom status modules

khsoh:
This feature improves the existing method of user custom modules by allowing the user to place all custom modules in a single directory.

Example: in the `.tmux.conf` file
```sh
# In ~/.tmux.conf, before loading the catppuccin plugin
set -gF @catppuccin_user_status_directory   "~/.config/tmux/custom"

# Apply your custom module in status-right
set -g status-right "#{E:@catpuccin_status_application}#{E:@catppuccin_status_cpu}#{E:@catppuccin_status_ram}"
```

Then create `~/.config/tmux/custom/ram.conf`:

```sh
%hidden MODULE_NAME="ram"

set -ogq @ram_low_fg_color    "#{E:@thm_fg}"
set -ogq @ram_medium_fg_color "#{E:@thm_fg}"
set -ogq @ram_high_fg_color   "#{E:@thm_crust}"

set -ogq @ram_low_bg_color    "#{E:@catppuccin_status_module_text_bg}"
set -ogq @ram_medium_bg_color "#{E:@catppuccin_status_module_text_bg}"
set -ogq @ram_high_bg_color   "#{E:@thm_red}"

set -ogq  "@catppuccin_${MODULE_NAME}_icon"           "#{l:#{ram_icon}} "
set -ogqF "@catppuccin_${MODULE_NAME}_color"          "#{E:@thm_blue}"
set -ogq  "@catppuccin_status_${MODULE_NAME}_text_fg" "#{l:#{ram_fg_color}}"
set -ogq  "@catppuccin_status_${MODULE_NAME}_text_bg" "#{l:#{ram_bg_color}}"
set -ogq  "@catppuccin_${MODULE_NAME}_text"           " #{l:#{ram_percentage}}"

source -F "#{@catppuccin_tmux_directory}/utils/status_module.conf"
```


kjnsn:
Thanks so much for working on this!

I'd like to understand better what makes this a better user experience. For example, I have a "custom" status line module in my config that is two lines:

```
set -g status-right "#[bg=#{@thm_flamingo},fg=#{@thm_crust}]#[reverse]#[noreverse]󱀙  "
set -ag status-right "#[fg=#{@thm_fg},bg=terminal] #(memory_pressure | awk '/percentage/{print $5}') "
```

---

> This feature improves the existing method of user custom modules by allowing the user to place all custom modules in a single directory.

Can you elaborate on this more? What about the plugin currently makes this more difficult?

I would like to keep things as simple as possible, and even avoid having separate config files at all.

khsoh:
I wrote this mainly for the [tmux-cpu]( https://git::@github.com/tmux-plugins/tmux-cpu) plugin.  So, this is a simple way to extend and add other configurations like cpu-temp, gpu, etc, in the status bar.

Its fine if this pull-request is not accepted in the interest of keeping this module simple.

Thank you for your consideration.

kjnsn:
> I wrote this mainly for the [tmux-cpu](https://git::@github.com/tmux-plugins/tmux-cpu) plugin. So, this is a simple way to extend and add other configurations like cpu-temp, gpu, etc, in the status bar.

How does it contrast with the existing method of using custom modules at https://github.com/catppuccin/tmux/blob/main/docs/tutorials/02-custom-status.md?

khsoh:
My intent was to keep the main .tmux.conf more focused on the "main" configuration and keep other status-related configuration in separate configuration files.

With the existing method of handling custom modules, it seems that I have to put all of the status-related configuration into the main configuration file.

PR #429, opened 2024-11-10
fix: specify the color for right separator of the status modules to be the same as text bg color

xafarr:
### Before
1. When a right separator for status module is specified then the color of the foreground is not same as the text background color.
2. There's no space between the windows when the status style is `rounded` making it look awkward

Screenshot 2024-11-11 at 9 17 52 am

### After the change

Screenshot 2024-11-11 at 10 00 45 am

### DIFF
--- a/catppuccin_options_tmux.conf
+++ b/catppuccin_options_tmux.conf
@@ -61,7 +61,7 @@ set -ogq @catppuccin_window_flags_icon_bell " 󰂞" # !
 set -ogq @catppuccin_window_flags_icon_format "##{?window_activity_flag,#{E:@catppuccin_window_flags_icon_activity},}##{?window_bell_flag,#{E:@catppuccin_window_flags_icon_bell},}##{?window_silence_flag,#{E:@catppuccin_window_flags_icon_silent},}##{?window_active,#{E:@catppuccin_window_flags_icon_current},}##{?window_last_flag,#{E:@catppuccin_window_flags_icon_last},}##{?window_marked_flag,#{E:@catppuccin_window_flags_icon_mark},}##{?window_zoomed_flag,#{E:@catppuccin_window_flags_icon_zoom},}"
 
 # Status line options
-set -ogq @catppuccin_status_left_separator " "
+set -ogq @catppuccin_status_left_separator ""
 set -ogq @catppuccin_status_middle_separator ""
 set -ogq @catppuccin_status_right_separator "█"
 set -ogq @catppuccin_status_connect_separator "yes" # yes, no
--- a/catppuccin_options_tmux.conf
+++ b/catppuccin_options_tmux.conf
@@ -63,7 +63,7 @@ set -ogq @catppuccin_window_flags_icon_format "##{?window_activity_flag,#{E:@cat
 # Status line options
 set -ogq @catppuccin_status_left_separator ""
 set -ogq @catppuccin_status_middle_separator ""
-set -ogq @catppuccin_status_right_separator "█"
+set -ogq @catppuccin_status_right_separator " "
 set -ogq @catppuccin_status_connect_separator "yes" # yes, no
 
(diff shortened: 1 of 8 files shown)


kjnsn:
I have a few questions here:

1. For the space between windows, what is wrong with `window-status-separator`?
2. This will break other people's configurations, can you achieve the same thing by modifying your own config?
3. Does setting `@catppuccin_status_right_separator "#[fg=#{@_ctp_module_text_bg}]\uE0B4"` not work?

xafarr:
Good questions. Let me answer respectively.

> For the space between windows, what is wrong with `window-status-separator`?

If you see the windows under `Before` heading are squished together and don't look as mentioned in the documentation, and also is different from how it looked previously. Please have a look at the attached screenshots below (look at the red circle).
### Actual
Screenshot 2024-11-11 at 9 17 52 am

### Documentation
Screenshot 2024-11-11 at 11 29 38 am


>  This will break other people's configurations, can you achieve the same thing by modifying your own config?

Can you please explain how will it break other people's configurations? It can be done but it's a too low-level configuration for the user in my opinion. At least it should look like the example configuration.

> Does setting `@catppuccin_status_right_separator` "#[fg=#{@_ctp_module_text_bg}]\uE0B4" not work?

It will work (haven't tried it) but isn't the expectation would be that the right separator should be of the same color as the text background. Same as how the left separator has the `fg` set, the right separator `fg` has to be set.

Screenshot 2024-11-11 at 11 44 00 am

kjnsn:
> Good questions. Let me answer respectively.
> 
> > For the space between windows, what is wrong with `window-status-separator`?
> 
> If you see the windows under `Before` heading are squished together and don't look as mentioned in the documentation, and also is different from how it looked previously.

Yep I can see that. If you add `set -g window-status-separator ' '` in your config does the same thing happen? See the philosophy doc on orthogonality of features; I don't want to add a feature that already exists in tmux itself.

> > This will break other people's configurations, can you achieve the same thing by modifying your own config?
> 
> Can you please explain how will it break other people's configurations? It can be done but it's a too low-level configuration for the user in my opinion. At least it should look like the example configuration.
> 
> > Does setting `@catppuccin_status_right_separator "#[fg=#{@_ctp_module_text_bg}]\uE0B4"` not work?
> 
> It will work (haven't tried it) but isn't the expectation would be that the right separator should be of the same color as the text background. Same as how the left separator has the `fg` set, the right separator `fg` has to be set.
> 

I made the decision to not do this to increase flexibility. You can have a space or text as the right separator if desired. This change means that those configs will break.

Tmux is generally quite difficult to configure and style. There are things like tmux-powerilne that make it easier at the cost of flexibility. I would prefer to expose only primitive styling in this plugin, with the hope that other things can be built on top of it. For example I'd love to see a "catppuccin powerline" repo somewhere that someone made

xafarr:
> Yep I can see that. If you add set -g window-status-separator ' ' in your config does the same thing happen? See the philosophy doc on orthogonality of features; I don't want to add a feature that already exists in tmux itself.

I think this makes sense and setting the `window-status-separator` works the same way. Good call. I'll remove this change.

> Does setting `@catppuccin_status_right_separator "#[fg=#{@_ctp_module_text_bg}]\uE0B4"` not work?

This doesn't work or may be I'm doing something wrong.

> I made the decision to not do this to increase flexibility. You can have a space or text as the right separator if desired. This change means that those configs will break.

Wouldn't you expect the same color for the separator even if it's a space? For the text something like `#[fg=#00000]Text` can be added as the separator value.

xafarr:
> Does setting `@catppuccin_status_right_separator "#[fg=#{@_ctp_module_text_bg}]\uE0B4"` not work?

It works if we make the following change in `status_module.conf`

Screenshot 2024-11-12 at 9 15 15 pm

xafarr:
If you have a disagreement with the proposed changes then please make the change in my last comment and I'll close this PR. Thanks a lot for your work.

kjnsn (approved):

PR #447, opened 2024-12-08
Inherit default style for popups

SfinxNT:
The current setup can cause style inconsistencies between popups and other windows/elements, particularly when using transparent backgrounds.

This change will resolve that issue (popups will inherit default colors for foreground and background).

Since this plugin doesn't utilize `window-active-style` or `menu-style`, there's no need for `popup-style` either.

### DIFF
--- a/catppuccin_tmux.conf
+++ b/catppuccin_tmux.conf
@@ -124,7 +124,6 @@ set -wgF pane-border-style "#{E:@catppuccin_pane_border_style}"
 
 # popups
 %if "#{>=:#{version},3.4}"
-  set -gF popup-style "bg=#{@thm_bg},fg=#{@thm_fg}"
   set -gF popup-border-style "fg=#{@thm_surface_1}"
 %endif
 


SfinxNT:
Before:
Screenshot_20241208_190420

After:
Screenshot_20241208_190527

btctrl (approved):
bump, who else needs to see this?

I came from trying to get tmux sessionx manager to be transparent

kjnsn:
I'm sympathetic to personal taste with regards to what colors things should be. At the same time the entire point of this repo is to be opinionated about what colors things should be in tmux. So I'm hesitant to remove colors and just use defaults, especially when catppuccin has a [style guide](https://github.com/catppuccin/catppuccin/blob/main/docs/style-guide.md) about what things like pop-ups should look like.

bl11y:
> I'm sympathetic to personal taste with regards to what colors things should be. At the same time the entire point of this repo is to be opinionated about what colors things should be in tmux. So I'm hesitant to remove colors and just use defaults, especially when catppuccin has a [style guide](https://github.com/catppuccin/catppuccin/blob/main/docs/style-guide.md) about what things like pop-ups should look like.

So how can I target the bg in my conf to use default?
Is there an option I'm missing somewhere?

kjnsn:
Just set the tmux option after you load catppuccin

611y:
> Just set the tmux option after you load catppuccin

I have this after running tpm
```# transparent background
set -g status-bg default
set -g status-style bg=default
```
but it doesn't target the floating pane bg, i guess it's something in sessionx that's the issue?

SfinxNT:
> > Just set the tmux option after you load catppuccin
> 
> I have this after running tpm
> 
> ```
> set -g status-bg default
> set -g status-style bg=default
> ```
> 
> but it doesn't target the floating pane bg, i guess it's something in sessionx that's the issue?

sessionx uses fzf under the hood to "draw" a popup window, fzf will inherit style from tmux for bg if it is not specified directly in fzf config.

So, for transparent bg in fzf/sessionx you need to set 
```tmux
set-option -wg popup-style bg=default
```

And this should be after catppuccin set it to `"bg=#{@thm_bg},fg=#{@thm_fg}"`. so, after running tpm

b11e:
> > > Just set the tmux option after you load catppuccin
> > 
> > 
> > I have this after running tpm
> > ```
> > set -g status-bg default
> > set -g status-style bg=default
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > but it doesn't target the floating pane bg, i guess it's something in sessionx that's the issue?
> 
> sessionx uses fzf under the hood to "draw" a popup window, fzf will inherit style from tmux for bg if it is not specified directly in fzf config.
> 
> So, for transparent bg in fzf/sessionx you need to set
> 
> ```shell
> set-option -wg popup-style bg=default
> ```
> 
> And this should be after catppuccin set it to `"bg=#{@thm_bg},fg=#{@thm_fg}"`. so, after running tpm

Thank you, this solved it

PR #465, opened 2025-01-18
fix(flags): remove additional space from flags

ItzYeho:
This effectively lets users decide if they want an additional space, instead of it being hard-coded

### DIFF
--- a/catppuccin_tmux.conf
+++ b/catppuccin_tmux.conf
@@ -168,7 +168,7 @@ set -ogqF @catppuccin_window_current_right_separator "#{@catppuccin_window_right
   set -gF window-status-bell-style "bg=#{@thm_yellow},fg=#{@thm_crust}"
 
    %if "#{==:#{@catppuccin_window_flags},icon}"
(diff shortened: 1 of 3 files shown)


catppuccin_tmux.conf
```diff
@@ -168,7 +168,7 @@ set -ogqF @catppuccin_window_current_right_separator "#{@catppuccin_window_right
   set -gF window-status-bell-style "bg=#{@thm_yellow},fg=#{@thm_crust}"
 
    %if "#{==:#{@catppuccin_window_flags},icon}"
-    set -gqF @_ctp_w_flags "#{E:@catppuccin_window_flags_icon_format} "
```
kjnsn, line 171: Please change the default value so that this doesn't effect existing users

kjnsn (changes requested):

catppuccin_tmux.conf
ItzYeho: Done, please check the renewed commit.

kjnsn (approved):

Issue #53, opened 2023-07-09
Tmux window names are not respected


aimestereo:
Notes: 
* I have latest version installed via TPM.
* I don't know if this behaviour by design, but at least it smth that should be configurable.

### Issue

If I rename window (or even pane), names in the bottom shows smth different (current folder?). 
For now I'm using this version that doesn't have this issue: https://github.com/dreamsofcode-io/catppuccin-tmux/commit/b4e0715356f820fc72ea8e8baf34f0f60e891718
Regretfully I can't be more useful, I don't even know if this version fixes smth, or if there's a bug in latest commits in main repo (https://github.com/dreamsofcode-io/catppuccin-tmux is quite behind main repo)

### Examples

<details>
  <summary>pic 1: main - wrong window names</summary>

  ![Screenshot 2023-07-09 at 12 23 22](https://github.com/catppuccin/tmux/assets/2869535/a06e7d21-cb3d-4347-89bd-1e8b2778eaf4)
</details>

<details>
  <summary>pic 2: alternative - window names are respected, if not set, shows current program name</summary>

  ![Screenshot 2023-07-09 at 12 24 40](https://github.com/catppuccin/tmux/assets/2869535/a3e629b3-2b3a-410f-92a1-6d23232bc28a)
</details>

---

### Additional info

My config, only relevant part:

```
set -g @plugin 'tmux-plugins/tpm'
set -g @plugin 'tmux-plugins/tmux-sensible'
#set -g @plugin 'dreamsofcode-io/catppuccin-tmux'  # catppuccin with more info in bottom line
set -g @plugin 'catppuccin/tmux'


# Initialize TMUX plugin manager (keep this line at the very bottom of tmux.conf)
run '~/.tmux/plugins/tpm/tpm'
```

My whole config
```
set -s mouse on

unbind C-b
set-option -g prefix M-a
bind-key M-a send-prefix

# Start windows and panes # with 1 (not 0), and renumber on close.
set -g base-index 1
set -g pane-base-index 1
set-option -g renumber-windows on

# set vi mode for selection and copy
set-window-option -g mode-keys vi
bind -T copy-mode-vi v send-keys -X begin-selection
bind -T copy-mode-vi y send-keys -X copy-pipe-and-cancel 'xclip -in -selection clipboard'

# forget the find window.  That is for chumps
bind-key -r f run-shell "tmux neww ~/.local/bin/tmux-sessionizer"


# Shift Alt vim keys to switch windows.
[...]

PR #472, opened 2025-01-30
docs: change recommended default config, so that window names are automatically set (#53)

jnussbaum:
The docs state that if the recommended default config is applied, tmux windows are automatically renamed according to the CWD. But as discussed in https://github.com/catppuccin/tmux/issues/53, this is currently not the case. This PR updates the recommended default config, to align it with what the docs say.

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,6 +160,13 @@ set -ag status-right "#{E:@catppuccin_status_session}"
 set -ag status-right "#{E:@catppuccin_status_uptime}"
 set -agF status-right "#{E:@catppuccin_status_battery}"
 
+# automatically set CWD as window name
+set -g @catppuccin_window_default_text "#W"
+set -g @catppuccin_window_current_text "#W"
+set -g @catppuccin_window_text "#W"
+set-option -g automatic-rename on
+set-option -g automatic-rename-format '#I #{b:pane_current_path}'
+
 run ~/.config/tmux/plugins/tmux-plugins/tmux-cpu/cpu.tmux
 run ~/.config/tmux/plugins/tmux-plugins/tmux-battery/battery.tmux
 # Or, if using TPM, just run TPM
--- a/docs/tutorials/01-getting-started.md
+++ b/docs/tutorials/01-getting-started.md
@@ -41,6 +41,13 @@ run ~/.config/tmux/plugins/catppuccin/tmux/catppuccin.tmux
 set -g status-left ""
 set -g status-right '#[fg=#{@thm_crust},bg=#{@thm_teal}] session: #S '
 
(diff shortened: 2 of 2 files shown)


kjnsn:
Can you point to where in the docs it mentions that the window names match the current directory?

README.md
```diff
@@ -160,6 +160,13 @@ set -ag status-right "#{E:@catppuccin_status_session}"
 set -ag status-right "#{E:@catppuccin_status_uptime}"
 set -agF status-right "#{E:@catppuccin_status_battery}"
 
+# automatically set CWD as window name
```
kjnsn, line 163: Why is `#W` the current directory?

README.md
```diff
@@ -160,6 +160,13 @@ set -ag status-right "#{E:@catppuccin_status_session}"
...
+set -g @catppuccin_window_default_text "#W"
```
kjnsn, line 164: This option no longer exists

README.md
```diff
@@ -160,6 +160,13 @@ set -ag status-right "#{E:@catppuccin_status_session}"
...
+set -g @catppuccin_window_current_text "#W"
+set -g @catppuccin_window_text "#W"
+set-option -g automatic-rename on
```
kjnsn, line 167: Why automatic rename if `#W` is used? It updates automatically anyway

And please be consistent with `set` and `set-option`. They are aliases and `set` is used everywhere else

README.md
```diff
@@ -160,6 +160,13 @@ set -ag status-right "#{E:@catppuccin_status_session}"
...
+set-option -g automatic-rename-format '#I #{b:pane_current_path}'
```
kjnsn, line 168: What does this do?

kjnsn (changes requested):

README.md
jnussbaum: I took this from https://github.com/catppuccin/tmux/issues/53#issuecomment-1675211611. But I'm not knowledgeable enough to say what it means.

README.md
jnussbaum: Sorry, took it from https://github.com/catppuccin/tmux/issues/53#issuecomment-1675211611.

jnussbaum:
> Can you point to where in the docs it mentions that the window names match the current directory?

There are 2 sources that imply that this would be the case:
- The screenshots in the Readme: All windows are named after the cwd/path to the cwd
- https://youtu.be/DzNmUNvnB04?t=626

But you're right that it is not stated explicitly in your docs. So perhaps this PR is obsolete.

reevik (approved):
It worked excellent!

PR #479, opened 2025-02-06
Make text fg color configurable again

nastye:
I recently updated and realized my config was incompatible and i wasn't able to reproduce my original config. I re-added the necessary options and glanced over docs to make sure there was some kind of hint at the options.

I will double check docs when i find a minute to do so :)


kjnsn:
Why doesn't `set -g @catppuccin_window_number "#[fg=#{@thm_fg},bg=#{@thm_crust}] #I"` work?

assets/structure.svg
kjnsn, line 1: Woah, what happened to the legibility?

catppuccin_options_tmux.conf
```diff
@@ -38,12 +38,16 @@ set -ogq @catppuccin_pane_number_position "left" # right, left
 
 # Window options
 set -ogq @catppuccin_window_status_style "basic" # basic, rounded, slanted, custom, or none
-set -ogq @catppuccin_window_text_color "#{@thm_surface_0}"
```
kjnsn, line 41: How will you maintain backwards compatibility?

assets/structure.svg
nastye: White background was removed (accidentally)

catppuccin_options_tmux.conf
nastye: current color does what bg would do for me, so set bg to color if unset before setting to the default if unset.

nastye:
> Why doesn't `set -g @catppuccin_window_number "#[fg=#{@thm_fg},bg=#{@thm_crust}] #I"` work?

works but wasn't documented and didn't think of it. I don't mind abandoning since technically a non-issue for me then.

kjnsn:
> > Why doesn't `set -g @catppuccin_window_number "#[fg=#{@thm_fg},bg=#{@thm_crust}] #I"` work?
> 
> 
> 
> works but wasn't documented and didn't think of it. I don't mind abandoning since technically a non-issue for me then.

It's basic tmux styling. Will close the PR then

PR #484, opened 2025-02-10
feat: Create a config linter

kjnsn:


validation/validate.py
```diff
@@ -0,0 +1,154 @@
...
+def print_error(message: str, line: str, lineno: int):
+    """Prints an error that occured on the given lineno, with the contents being line."""
+    print(f"{message}, at line {lineno}: {line}")
+    exit(1)
+
+
+def parse_config(config: str, verbose: bool) -> list[TmuxCommand]:
+    commands: list[TmuxCommand] = []
+    # TODO: Handle multiline commands that contain '\'
+    for line_number, line in enumerate(config.splitlines()):
```
backwardspy, line 61:
```suggestion
    for line_number, line in enumerate(config.splitlines(), start=1):
```
to start counting lines at 1 rather than 0, unless i've missed a +1 somewhere else!

validation/validate.py
```diff
@@ -0,0 +1,154 @@
...
+        cmd_re = r"\s*(\w+)\s+(-\w+)?\s*(@?[\w\-_]+)?\s*[\"']?(.*)"
+        m = re.match(cmd_re, line)
```
backwardspy, line 63: you can `re.compile` this pattern once at the beginning and then match with that, rather than recompiling on each iteration of the loop

backwardspy (commented):
nice work! just a couple of suggestions to consider.

validation/validate.py
backwardspy: it might not be. in theory, compiling it once and then using it many times should be kinder on system resources. however, i believe python maintains its own cache for compiled regex patterns anyway, so you may see no difference at all.
i'm also not sure how long the average tmux config is; if this loop only runs a few hundred times it would be a negligible difference.
thought it was worth pointing out for consideration regardless.

kjnsn: It is cached, and I'm not really interested in micro optimisations without a measurable performance benefit that users would notice

backwardspy: as i said, just a suggestion :)

PR #491, opened 2025-02-19
feat: change color scheme to gruvbox

filipcujanovic:


kjnsn:
What I find really funny is that you can override any colors by setting them like `set -g @thm_fg "#FFFFFF"`, so creating a fork is so much extra effort

filipcujanovic:
@kjnsn Found out too late for that plus I wanted to make a gruvbox theme based on the catppuccin base 😅

PR #492, opened 2025-02-22
docs: Correct the plugin path for TPM in the readme

### LINKED ISSUE #476: Incorrect Installation Instructions for TPM in Documentation
### Is there an existing issue outlining your problem?

- [x] I have searched the existing issues and they do not solve my problem.

### Describe your issue.

The installation instructions for using TPM (Tmux Plugin Manager) with Catppuccin appear to be incorrect. The README currently suggests the following path: `~/.config/tmux/plugins/tmux/catppuccin.tmux` .
However, on my system, this path did not exist. Instead, I was able to successfully load Catppuccin using: `~/.tmux/plugins/tmux/catppuccin.tmux`.

## Expected Behavior
The documentation should provide an accurate default path for TPM installations or clarify potential variations based on different setups.

## Suggested Fix
* Update the README to reflect the correct default path.
* If paths can vary depending on configuration, mention this in the documentation and suggest checking ~/.tmux/plugins/ as a common alternative.
<hr />

Let me know if you'd like any additional details or testing on this!

ThomasCode92:
Correct the plugin path for TPM in the readme. Fixes #476 

[TPM](https://github.com/tmux-plugins/tpm) recommends the directory `~/.tmux/plugins/tpm` be used as the installation directory in the README, which is likely what most users will choose. TPM can be installed into _any_ directory, but the readme should ideally mention the most common choice that users will have, which we can assume is going to be `~/.tmux/plugins`.

---

Tested this change on my  [current setup](https://github.com/ThomasCode92/dotfiles/blob/main/dot_tmux.conf), see screenshot below.

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,7 +151,7 @@ set -g @catppuccin_window_status_style "rounded"
 
 # Load catppuccin
 run ~/.config/tmux/plugins/catppuccin/tmux/catppuccin.tmux
-# For TPM, instead use `run ~/.config/tmux/plugins/tmux/catppuccin.tmux`
+# For TPM, instead use `run ~/.tmux/plugins/tmux/catppuccin.tmux`
 
 # Make the status line pretty and add some modules
 set -g status-right-length 100


kjnsn:
I'm going to have to rewrite your PR title and description. The title should describe what the change is. "Update README.md for TPM" doesn't tell me what the change is in enough detail, but "Correct the plugin path for TPM in the readme" does.

Additionally I'm very surprised you didn't reference https://github.com/tmux-plugins/tpm, explaining why you think `.config` was added and why removing it is the right choice, not just for your machine but for the thousands of people using this. When it comes to public documentation it doesn't matter if it works on your machine so much as demonstrating it will work for everyone else as well.

kjnsn (approved):

ThomasCode92:
Ok, thanks for the reply. Will keep that in mind for the next time!

PR #500, opened 2025-03-05
Add a "reset" option that will reset all options when loading the plugin

kjnsn:
This only works with a super recent tmux change: https://github.com/tmux/tmux/pull/4353

Will document how it works when the next major version of tmux is released

### DIFF
--- a/catppuccin_options_tmux.conf
+++ b/catppuccin_options_tmux.conf
@@ -3,6 +3,80 @@
 # This is executed separately to the main configuration
 # so that options are set before parsing the rest of the config.
 
+# Reset everything if requested.
+#
+# Useful for auto switching between themes:
+#
+# set-hook -g client-dark-theme {
+#   set -g @catppuccin_flavor "frappe"
+#   set -g @catppuccin_reset "true"
+#   run ~/code/github.com/catppuccin/tmux/catppuccin.tmux
+# }
+# set-hook -g client-light-theme {
+#   set -g @catppuccin_flavor "latte"
+#   set -g @catppuccin_reset "true"
+#   run ~/code/github.com/catppuccin/tmux/catppuccin.tmux
+# }
+%if "#{==:#{@catppuccin_reset},true}"
+set -Ugq @thm_bg
+set -Ugq @thm_fg
+set -Ugq @thm_rosewater
+set -Ugq @thm_flamingo
+set -Ugq @thm_rosewater
+set -Ugq @thm_pink
+set -Ugq @thm_mauve
+set -Ugq @thm_red
+set -Ugq @thm_maroon
+set -Ugq @thm_peach
+set -Ugq @thm_yellow
+set -Ugq @thm_green
+set -Ugq @thm_teal
+set -Ugq @thm_sky
+set -Ugq @thm_sapphire
+set -Ugq @thm_blue
+set -Ugq @thm_lavender
+set -Ugq @thm_subtext_1
+set -Ugq @thm_subtext_0
+set -Ugq @thm_overlay_2
+set -Ugq @thm_overlay_1
+set -Ugq @thm_overlay_0
+set -Ugq @thm_surface_2
+set -Ugq @thm_surface_1
+set -Ugq @thm_surface_0
+set -Ugq @thm_mantle
+set -Ugq @thm_crust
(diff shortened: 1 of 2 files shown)


catppuccin_options_tmux.conf
```diff
@@ -3,6 +3,80 @@
...
+# set-hook -g client-dark-theme {
+#   set -g @catppuccin_flavor "frappe"
+#   set -g @catppuccin_reset "true"
+#   run ~/code/github.com/catppuccin/tmux/catppuccin.tmux
+# }
+# set-hook -g client-light-theme {
+#   set -g @catppuccin_flavor "latte"
+#   set -g @catppuccin_reset "true"
+#   run ~/code/github.com/catppuccin/tmux/catppuccin.tmux
+# }
```
rogeruiz, line 19: Great work on this. That patch you linked to is exciting. I'm not sure when it's planned for release though. Do you think these should be configurable so users can set the path and the flavor? I wonder if we should document the `set-hook` functionality instead of incorporating it. But I'm also not super familiar with the feature yet.

catppuccin_options_tmux.conf
```diff
@@ -3,6 +3,80 @@
...
+%if "#{==:#{@catppuccin_reset},true}"
```
rogeruiz, line 20: so elegant! The `-U` flag is interesting. I didn't know about that flag.

rogeruiz (commented):
lgtm 🌈

I like the direction of this a lot @kjnsn. Thanks for getting started on this work.

> I'll be waiting to actually approve this when the work is all complete. but it's a good start 🚀

catppuccin_options_tmux.conf
kjnsn: What should be configurable? These are comments

Also hooks are already documented, unless there's something special here worth mentioning? https://man7.org/linux/man-pages/man1/tmux.1.html#HOOKS

rogeruiz: Ah, yeah I think I missed that the first time. I think the example can stay as a comment then.

dimitur2204:
Hey! Is the PR in a working state? I would like to try and run that locally for me.

rogeruiz:
Sure @dimitur2204 its stable enough to test but it's not fully documented. Take a look at this Nix derivation below to write a shell script that resets the flavor from Latte to Mocha. Keep in mind it's my personal setup but the steps in the script should help make it clear for you. So you will need to customize things based on the documentation on this branch.

We don't currently have a timeline for when this will make it into a release. Testing is always appreciated it.

➡️ https://github.com/rogeruiz/srht-dotfiles.nix/blob/main/module/tools/terminal/tmux/catppuccin/bin/default.nix

dimitur2204:
I just managed to get it to work on MacOS with [dark-notify](https://github.com/erikw/tmux-dark-notify). This is what I have in the `.tmux.conf`:
```
set -g @catppuccin_reset "true"
run ~/.tmux/plugins/catppuccin/tmux/catppuccin.tmux
set -g @plugin 'erikw/tmux-dark-notify'
set -g @dark-notify-theme-path-light '~/.tmux/plugins/catppuccin/tmux/themes/catpuccin_set_latte.conf'
set -g @dark-notify-theme-path-dark '~/.tmux/plugins/catppuccin/tmux/themes/catpuccin_set_mocha.conf'
```

And those "catpuccin_set_{theme}.conf" are those snippets from the suggested documentation:
```
set -g @catppuccin_flavor "{theme}"
set -g @catppuccin_reset "true"
run ~/.tmux/plugins/catppuccin/tmux/catppuccin.tmux
```

Lovely work guys!

rogeruiz:
Thanks @dimitur2204 for the feedback. Read through the docs on this branch https://github.com/catppuccin/tmux/pull/512. There's a section about why certain background colors need to be set even though we don't normally ask users to set them.

djensenius:
Hi @rogeruiz et al, wondering where this is and if you need any (non-LLM) help getting it across the finish line.

rogeruiz:
Thanks @djensenius for the question. I have to dust off my local copy and I'll report back this weekend. It should be fairly easy to update. We may just need documentation around it really. But I'll need to take a closer look.

PR #502, opened 2025-03-09
feat: add network bandwidth status panel

matawed:


kjnsn:
Hey! Thanks for your hard work on this, I can see you've gone to a bunch of effort. Unfortunately I'm going to have to reject this for two reasons:

1. We are not adding any more status line modules. User's can add their own trivially, and each one adds a maintenance burden.
2. The scripts you linked (https://github.com/ilya-manin/tmux-network-bandwidth/blob/63c6b3283d537d9b86489c13b99ba0c65e0edac8/scripts/network-bandwidth.sh) have not been updated in over five years, and use netstat which is very deprecated on linux.

Feel free to add some documentation on adding a DIY section to the status line for bandwidth though :)

PR #503, opened 2025-03-09
fix: empty load status panel

matawed:
without PR:
image

with PR:
image

env:

tmux 3.4
tmux-network-bandwidth: 
https://github.com/jamesoff/tmux-loadavg/commit/91319eff74ee677efb77c882dcc8e3b8780dc3a2
catppuccin: https://github.com/catppuccin/tmux/commit/073ee54992c59fedcc29c1525a26f95691f0ae1f

### DIFF
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -9,5 +9,6 @@ script_dir=$(cd "$(dirname "${BASH_SOURCE[0]}")" &>/dev/null && pwd -P)
 
 "${script_dir}"/tests/harness.sh --test "${script_dir}"/tests/application_module.sh --expected "${script_dir}"/tests/application_module_expected.txt "$@"
 "${script_dir}"/tests/harness.sh --test "${script_dir}"/tests/battery_module.sh --expected "${script_dir}"/tests/battery_module_expected.txt "$@"
+"${script_dir}"/tests/harness.sh --test "${script_dir}"/tests/load_module.sh --expected "${script_dir}"/tests/load_module_expected.txt "$@"
 "${script_dir}"/tests/harness.sh --test "${script_dir}"/tests/cpu_module.sh --expected "${script_dir}"/tests/cpu_module_expected.txt "$@"
 "${script_dir}"/tests/harness.sh --test "${script_dir}"/tests/pane_styling.sh --expected "${script_dir}"/tests/pane_styling_expected.txt "$@"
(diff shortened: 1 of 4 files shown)


kjnsn:
I'd prefer to invest in getting rid of an external dependency and just using this one-liner instead: https://github.com/tmux/tmux/wiki/FAQ#what-is-the-best-way-to-display-the-load-average-why-no-l

I actually don't know why https://github.com/jamesoff/tmux-loadavg even exists. In fact we shouldn't be using it at all since it hardcodes color values that are not catppuccin: https://github.com/jamesoff/tmux-loadavg/blob/91319eff74ee677efb77c882dcc8e3b8780dc3a2/scripts/display_load.sh#L38

matawed:
Ok, I agree. Makes sense....

However, I actually started yesterday with putting https://github.com/catppuccin/tmux/blob/main/docs/tutorials/02-custom-status.md in my `.tmux.conf` so I had something like the following

```
## catppuccin/tmux
set -g @catppuccin_flavor 'mocha'
set -g @catppuccin_window_current_text "#{window_name}"
set -g @catppuccin_window_text "#{window_name}"
# the following is needed as tpm only runs *.tmux in  ~/.tmux/plugins/*/
run ~/.tmux/plugins/catppuccin/tmux/catppuccin.tmux


set -g status-right-length 100
set -g status-left-length 100
set -g status-left ""
set -g status-right "#{E:@catppuccin_status_directory}"
#set -agF status-right "#{E:@catppuccin_status_load_new}"
#set -agF status-right "#{E:@catppuccin_status_cpu}"
#set -agF status-right "#{E:@catppuccin_status_network_bandwidth}"
# In ~/.tmux.conf, after the catppuccin plugin has been loaded.


%hidden MODULE_NAME="load_new"
set -ogq "@catppuccin_${MODULE_NAME}_icon" " "
set -ogqF "@catppuccin_${MODULE_NAME}_color" "#{E:@thm_pink}"
set -ogq "@catppuccin_${MODULE_NAME}_text" "#(uptime|awk '{split(substr($0, index($0, "load")), a, ":"); print a[2]})'"
source "~/.tmux/plugins/catppuccin/tmux/utils//status_module.conf"

set -ag status-right "#{E:@catppuccin_status_load_new}"
```

But this unfortunately doesn't work. I also tried with hardcoding MODULE_NAME directly in the set statements and also with a hardcoded `_text`  but I couldn't get it to work. 

What did I do wrong?

Issue #527, opened 2025-04-22
status modules like "directory" or "application" are not updated (+FIX)


itakatz:
### I am not requesting help with my configuration and believe something is genuinely broken.

- [x] Something is really actually broken.

### Is there an existing issue outlining your problem?

- [x] I have searched the existing issues and [troubleshooting guide](https://github.com/catppuccin/tmux/blob/main/docs/guides/troubleshooting.md), and they do not solve my problem.

### Describe your problem.

Using tag `v2.1.3`. 
I set the status message using for example:

`set -gF status-right "#{E:@catppuccin_status_directory}#{E:@catppuccin_status_application}#{E:@catppuccin_status_date_time}"`

however the current directory and application are not updated in the status.

[Possible fix] I believe the reason is that in this line:
https://github.com/catppuccin/tmux/blob/14a546fb64dc1141e5d02bac2185d8c1fd530d6a/utils/status_module.conf#L42
the `@catppuccin_${MODULE_NAME}_text` variable is expanded (using the `E:`), so it no longer includes the tmux vars like `pane_current_path` or `pane_current_command` etc.
Removing the `E:` seems to fix it, that is:
`set -ag "@catppuccin_status_${MODULE_NAME}" "#{@catppuccin_${MODULE_NAME}_text}"`

(PS sorry for not fixing it using a PR)

### Paste your configuration.

set -g @catppuccin_flavor 'mocha' #'latte' # or frappe, macchiato, mocha

set -g @catppuccin_window_flags "icon" # "text"
set -g @catppuccin_window_status_style "rounded"
set -g @catppuccin_window_current_number_color "#{@thm_mauve}"

set -g @catppuccin_window_left_separator ""
set -g @catppuccin_window_right_separator " "
set -g @catppuccin_window_middle_separator " █"
set -g @catppuccin_window_number_position "right"

#set -g @catppuccin_window_default_fill "number"
set -g @catppuccin_window_text "#W"

#set -g @catppuccin_window_current_fill "number"
set -g @catppuccin_window_current_text "#W"

set -g @catppuccin_status_left_separator  " "
set -g @catppuccin_status_right_separator ""
set -g @catppuccin_status_right_separator_inverse "no"
set -g @catppuccin_status_connect_separator "no"

set -g @catppuccin_directory_text "#{pane_current_path}"
set -g @catppuccin_application_text "#{pane_current_command}"

run ~/.config/tmux/plugins/catppuccin/tmux/catppuccin.tmux

[...]

PR #528, opened 2025-04-23
Update status-line.md

itakatz:
Add remark about using `-F` option for `set` when setting the status.
See https://github.com/catppuccin/tmux/issues/527


docs/reference/status-line.md
```diff
@@ -14,6 +14,7 @@ set -g status-right-length 100
 set -g status-right "#{E:@catppuccin_status_application}#{E:@catppuccin_status_session}"
 set -g status-left ""
 ```
+Note that the status variables are expanded using `E:` before the variable name. Depending on the module used, you may also have to use `set -gF` in order for `set` to expand format, see for example the [CPU](#cpu-module) and [battery](#battery-module) modules below.
```
kjnsn, line 17: nit: tmux calls them "options", so "Note that options are expanded as format strings by placing `E:` before the option name".

kjnsn, line 17: `set -gF` worth being specific that `F` is the flag that treats the value passed as a format string that is immediately expanded. Also worth linking to the man page

kjnsn, line 17: The status line expands format strings in `status-{left,right}` anyway before displaying them, so worth explaining why some modules need to their formats fully expanded prior to being added to the `status-{left,right}` options

kjnsn (changes requested):

docs/reference/status-line.md
itakatz: done

rogeruiz (approved):
lgtm 🌈

Good work on this @itakatz. Your updates address @kjnsn's review comments.

Issue #459, opened 2025-01-09
Kube module blank


mhkarimi1383:
### I am not requesting help with my configuration and believe something is genuinely broken.

- [X] Something is really actually broken.

### Is there an existing issue outlining your problem?

- [X] I have searched the existing issues and they do not solve my problem.

### Describe your problem.

I have search for similar problems but it was closed,

My kube-module is blank :(

### Paste your configuration.

See also github.com/mhkarimi1383/tmux.git and github.com:mhkarimi1383/nixos.git

```
set -g mouse on
set -g default-terminal "tmux-256color"
set -g @catppuccin_flavor 'mocha'
set -g @catppuccin_window_status_style 'rounded'

set -g @catppuccin_kube_context_color "#{@thm_red}"
set -g @catppuccin_kube_namespace_color "#{@thm_sky}"

run ~/.config/tmux/plugins/catppuccin/tmux/catppuccin.tmux

set -g status-right-length 100
set -g status-left-length 100
set -g status-left ""
set -g status-right "#{E:@catppuccin_status_application}"
set -agF status-right "#{E:@catppuccin_status_cpu}"
set -ag status-right "#{E:@catppuccin_status_session}"
set -ag status-right "#{E:@catppuccin_status_uptime}"
set -agF status-right "#{E:@catppuccin_status_battery}"
set -agF status-right "#{@catppuccin_status_gitmux}"
set -agF status-right "#{E:@catppuccin_status_kube}"

set -g @plugin 'tmux-plugins/tpm'
set -g @plugin 'tmux-plugins/tmux-sensible'
set -g @plugin 'tmux-plugins/tmux-battery'
set -g @plugin 'tmux-plugins/tmux-cpu'
set -g @plugin 'jonmosco/kube-tmux'
run '~/.tmux/plugins/tpm/tpm'
```

### Attach screenshots.

![image](https://github.com/user-attachments/assets/97bc8df2-5a45-43bc-bfd7-a809f5304a31)

### What tmux version are you seeing the issue on?

tmux 3.5a

mhkarimi1383:
Also `rounded style is not applied to the left side and right size is not applied too

kjnsn:
What is the output of `tmux show -g status-right`? It's also possible that you need to increase the length of status-right since you've got so much stuff, can you try with something like 200?

mhkarimi1383:
```shell
tmux show -g status-right
[...]

PR #531, opened 2025-05-08
fix(status): change tmux plugin for kube module

dhollinger:
Fixed the kube module by doing the following:

related to #459

* Replaced `jonmosco/kube-tmux` with `tony-sol/tmux-kubectx` as the
former was incorrectly formatting the tmux output and doesn't look to be
maintained at all anymore.
* Updated the Kube Module's documentation to account for the change in
tmux plugin being used.

Example: 
image

### DIFF
--- a/docs/reference/status-line.md
+++ b/docs/reference/status-line.md
@@ -220,9 +220,9 @@ run '~/.tmux/plugins/tpm/tpm'
 
 ## Kube module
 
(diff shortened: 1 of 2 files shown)


rogeruiz (approved):
lgtm 🌈

Thanks @dhollinger for these changes. I don't have a k8s cluster to test this against, but the documentation you revised and the updated module name text look accurate to me. Thanks for providing a screenshot as well.

Issue #409, opened 2024-10-23
Transparency 


kjnsn:
### Is there an existing issue outlining your problem?

- [X] I have searched the existing issues and they do not solve my problem.

### Describe your issue.

transparency

PR #577, opened 2025-12-31
Fix: Transparent status background support with rounded and slanted styles

jerek:
This is actually two fixes, which address [issue #409](https://github.com/catppuccin/tmux/issues/409). This PR is also an alternative to [pr #568](https://github.com/catppuccin/tmux/pull/568).
* Always use the status bg color as an actual background color (as opposed to a foreground color), so that terminals can make its color transparent and it works in rounded and slanted styles.
* Set the default bg color when `@catppuccin_status_background` is set to 'none', so that it works in rounded and slanted styles.

### Details

The [configuration documentation](https://github.com/catppuccin/tmux/blob/8b0b9150f9d7dee2a4b70cdb50876ba7fd6d674a/docs/reference/configuration.md#status-line) says to set `@catppuccin_status_background` to `none` to make the status bar transparent, but this doesn't work in the rounded or slanted window status styles. Attempting to work around the problem by setting it to the background color that the terminal is making transparent also doesn't work for the rounded or slanted styles.

These two similar issues have different causes:
1. While terminals can have a particular color chosen as the one to make transparent, it only works on background colors, and the existing configuration was using the status background color as a foreground color.
    * This PR's fix: Change the usage of `@catppuccin_status_background`'s value to only apply it as a background color, which allows users to set it to a background color that their terminal will make transparent.
2. Setting `@catppuccin_status_background` to `none` doesn't work due to how it internally sets `@_ctp_status_bg` to 'none', which isn't valid for tmux. In the tmux man page the value 'none' is referenced for clearing all formatting, but not as a valid value for a particular property like `bg` or `fg`. Instead, the man page has an example of 'default' being used for that.
    * This PR's fix: When the user sets `@catppuccin_status_background` to 'none', set the value of `@_ctp_status_bg` to 'default' instead of 'none', which tmux handles correctly.

I tested the changes in Wezterm primarily, but also checked that it works in Kitty and Ghostty.

### Comparison Screenshots

**Before** — background set to 'none'
image

**Before** — background set to terminal's transparent background color
image

**After** — either transparency method
image

**After** — non-transparent colors still work as expected, as shown here
image

### DIFF
--- a/catppuccin_tmux.conf
+++ b/catppuccin_tmux.conf
@@ -136,29 +136,38 @@ set -wgF pane-border-style "#{E:@catppuccin_pane_border_style}"
 
 %elif "#{==:#{@catppuccin_window_status_style},rounded}"
 
-  set -gq @catppuccin_window_left_separator "#[fg=#{@catppuccin_window_number_color},bg=#{@_ctp_status_bg}]#[fg=#{@catppuccin_window_text_color},bg=#{@catppuccin_window_number_color}]"
-  set -gq @catppuccin_window_current_left_separator "#[fg=#{@catppuccin_window_current_number_color},bg=#{@_ctp_status_bg}]#[fg=#{@catppuccin_window_current_text_color},bg=#{@catppuccin_window_current_number_color}]"
   set -gq @catppuccin_window_middle_separator " "
-  set -gq @catppuccin_window_right_separator "#[fg=#{@catppuccin_window_text_color},bg=#{@_ctp_status_bg}]#[none]"
-  set -gq @catppuccin_window_current_right_separator "#[fg=#{@catppuccin_window_current_text_color},bg=#{@_ctp_status_bg}]#[none]"
+  %if "#{==:#{@catppuccin_window_number_position},left}"
+    set -gq @catppuccin_window_left_separator "#[fg=#{@catppuccin_window_number_color},bg=#{@_ctp_status_bg}]#[fg=#{@catppuccin_window_text_color},bg=#{@catppuccin_window_number_color}]"
+    set -gq @catppuccin_window_current_left_separator "#[fg=#{@catppuccin_window_current_number_color},bg=#{@_ctp_status_bg}]#[fg=#{@catppuccin_window_current_text_color},bg=#{@catppuccin_window_current_number_color}]"
+    set -gq @catppuccin_window_right_separator "#[fg=#{@catppuccin_window_text_color},bg=#{@_ctp_status_bg}]#[none]"
+    set -gq @catppuccin_window_current_right_separator "#[fg=#{@catppuccin_window_current_text_color},bg=#{@_ctp_status_bg}]#[none]"
+  %else
+    set -gq @catppuccin_window_left_separator "#[fg=#{@catppuccin_window_text_color},bg=#{@_ctp_status_bg}]#[fg=#{@catppuccin_window_text_color},bg=#{@catppuccin_window_number_color}]"
+    set -gq @catppuccin_window_current_left_separator "#[fg=#{@catppuccin_window_current_text_color},bg=#{@_ctp_status_bg}]#[fg=#{@catppuccin_window_current_text_color},bg=#{@catppuccin_window_current_number_color}]"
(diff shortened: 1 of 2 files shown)


retroquake:
Hi! I tried using this branch and the colors are weird now.

issue

Here's my config:

```
set -g @catppuccin_flavor "mocha"
set -ogq @catppuccin_window_flags "icon"
set -ogq @catppuccin_window_text " #W"
set -ogq @catppuccin_window_current_text " #W"
set -ogq @catppuccin_uptime_icon " "
set -g @catppuccin_status_background "none"

source -F "~/.config/tmux/custom_modules/time.conf"

# Clear the status lines to prevent ghost-appends on reload
set -g status-left ""
set -g status-right ""

set -ag status-right "#{E:@catppuccin_status_session}"
set -ag status-right "#{E:@catppuccin_status_uptime}"
set -ag status-right "#{E:@catppuccin_status_time}"

```

jerek:
Ah, yeah, the whole "none" vs "default" thing is a bit confusing. It seems like the usage of "none" here is a keyword specific to the Catppuccin theme for Tmux, which activates its transparency features.

Maybe "transparent" would be clearer? Though that's outside the scope of this fix anyway.

phucisstupid:
> @Pawel-Parma Fixed!
> 
> image

how you config tmux to show both directory name and running program?

jerek:
> how you config tmux to show both directory name and running program?

It's a combination of things. I don't change the settings of tmux itself for this, but rather the settings of zsh (which I use instead of bash) and Neovim. I also customize the tmux pane title in various scripts of mine.

Here are some of my settings:

<details>
  <summary>In my zshrc</summary>

```zsh
autoload -Uz add-zsh-hook

_precmd_title() {
    print -Pn "\e]0;%~\a"
}

add-zsh-hook precmd _precmd_title
```
And if you want to add something before the file path, such as a [Nerd Font](https://www.nerdfonts.com/) folder icon, change the function in the middle like this:
```zsh
_precmd_title() {
    print -Pn "\e]0; %~\a"
}
```
</details>

<details>
  <summary>In my Neovim Lua config</summary>

- The two `local` vars at the top are an easy way to customize this a little.
- I commented out the path substitutions section, since that calls one of my custom scripts. But I left it in a comment, in case you want to do something like that.

```lua
--  TITLE 
function _G.getTitleString()
    local prefix = ' '
    local maxChars = 50

    -- Get the name of the current buffer.
    local name = vim.api.nvim_buf_get_name(0)

    -- If there's no open buffer, just show the Neovim name.
    if name == '' then
        return prefix .. 'Neovim'
    end

    -- Use my path substitutions script to customize the name.
    -- local handle = io.popen(vim.fn.expand '~/.config/j-scripts/shell/path-substitutions.sh' .. ' "' .. prefix .. name .. '"')
    -- if handle then
    --     name = handle:read '*a'
    --     handle:close()
    -- end

    -- If the text is longer than maxChars, trim the left side.
    local substr = string.sub(name, -1 * maxChars)
    if string.len(substr) == maxChars then
        if string.sub(name, 0, 1) == '/' then
            substr = string.sub(name, -1 * (maxChars - 1))
        end

        local firstSlash = string.find(substr, '/')
        if firstSlash then
            substr = '…' .. string.sub(substr, firstSlash)
        end
    end

    return substr
end
vim.opt.title = true
vim.opt.titlelen = 0 -- Disable's Neovim's path limit, since it's handled above.
vim.opt.titlestring = '%{luaeval("getTitleString()")}'
```
</details>

<details>
  <summary>In my zsh/bash aliases</summary>

- This falls back on setting the title in WezTerm, but it won't error if you're using a different terminal emulator.
- I use this so that in my scripts I can include commands like:
`setSmartTitle '  Starting container...'`

```bash
# Sets the title of the tmux pane (tab), or the terminal when not in tmux.
setSmartTitle() {
    title=$*

    if [[ -z "$TMUX" || $TERM_PROGRAM == 'tmux' ]]; then
        tmux select-pane -T "$title"
    elif [[ $TERM_PROGRAM == 'WezTerm' ]]; then
        wezterm cli set-tab-title "$title"
    fi
}
```
</details>

PR #597, opened 2026-06-01
fix: escape hex colors so they survive tmux next-3.7 format expansion

### LINKED ISSUE #596: Theme colors break on tmux next-3.7: leading hex digit eaten by format expansion
### I am not requesting help with my configuration and believe something is genuinely broken.

- [X] Something is really actually broken.

### Is there an existing issue outlining your problem?

- [X] I have searched the existing issues and troubleshooting guide, and they do not solve my problem.

### Describe your problem.

On `tmux next-3.7`, every theme color whose **first hex digit is `1`–`9`** is corrupted at render time — the leading digit is dropped (`#15232d` → `5232d`), producing an invalid color. Status modules, the window list, and pane borders render with broken/transparent colors. Colors starting with `0` or `a`–`f` are unaffected.

**Root cause (upstream tmux):** tmux/tmux#5121 (the run-shell argument feature, merged 2026-06-01) added `#1`–`#9` as positional-argument shorthand directly in `format_expand1`. Any format expansion now treats a literal `#` followed by `1`–`9` as an (empty) positional argument and consumes it.

**Why it hits catppuccin:** the plugin resolves theme colors through `#{E:…}`, and the status line is rendered through `#{E:@catppuccin_status_*}`. So literal hex such as mocha's `@thm_bg #1e1e2e` / `@thm_crust #11111b` loses its leading digit. Every flavor has such colors, so all flavors are affected — not a single custom theme.

**Expected:** colors render intact (`#1e1e2e`), as on tmux ≤ 3.5a.

### Paste your configuration.

Minimal reproduction — **tmux only, no plugin needed** (proves it is the tmux change):

```sh
tmux -f /dev/null new-session -d
for d in 0 1 2 3 4 5 6 7 8 9 a b c d e f; do
  tmux set -g @c "#${d}3549f"
  printf '#%s3549f -> %s\n' "$d" "$(tmux display -p '#{E:@c}')"
done
tmux kill-server
```

On `next-3.7` this prints `#13549f -> 3549f` … `#93549f -> 3549f` (digits 1–9 eaten); on 3.5a all sixteen round-trip intact.

Plugin reproduction (default mocha):

```sh
set -g @catppuccin_flavor "mocha"
set -g status-right "#{E:@catppuccin_status_host}"
run ~/.config/tmux/plugins/catppuccin/tmux/catppuccin.tmux
```

The host module's crust foreground renders as `1111b` instead of `#11111b`.

### Attach screenshots.

Window/status separators render transparent or garbled wherever a `#1`-leading surface color is used.
|tmux next-3.7|expected state|
|--------------|---------------|
|Image|Image|

### What tmux version are you seeing the issue on?

tmux next-3.7

### What shell are you using?

bash / zsh (shell-independent)

### Any additional comments?

Brand-new: tmux/tmux#5121 merged 2026-06-01, so released tmux (≤ 3.6) is unaffected — only `next-3.7` / master builds are. A fix is prepared (resolve theme colors with `#{@thm_*}` instead of `#{E:@thm_*}` at build time, and escape resolved `#RRGGBB` → `##RRGGBB` in the render-expanded formats so the render unescapes them); PR to follow.

hoblin:
Closes #596.

## Problem

tmux `next-3.7` (via the run-shell argument feature, tmux/tmux#5121, merged 2026-06-01) added `#1`–`#9` as positional-argument shorthand in `format_expand1`. Expanding a format at **render time** now eats the first digit of any literal `#RRGGBB` whose first hex digit is 1–9 (`#15232d` → `5232d`), producing invalid colors.

Every flavor uses such colors (mocha `#1e1e2e` / `#11111b`, surfaces, …), so **status modules, the window list, and pane borders all render broken** under next-3.7. Released tmux ≤ 3.6 is unaffected.

## Fix

Two changes, one render-safety rule:

1. **Resolve theme colors with `#{@thm_*}` instead of `#{E:@thm_*}`** where the value is a literal hex. The extra `#{E:}` pass re-scans the resolved hex and eats the leading digit at *build* time; a single resolution doesn't. (`@thm_*` always holds a literal hex across every flavor, never a reference.)

2. **Escape `#RRGGBB` → `##RRGGBB`** (in `catppuccin.tmux`) for the formats tmux expands at *render* time — `window-status-format`, `window-status-current-format`, `pane-border-format`, and the `@catppuccin_status_*` module strings. The render unescapes them back to intact `#RRGGBB`. Deferred references and runtime conditionals carry no literal and are untouched. The pass is **idempotent** (unescape-then-escape), so it survives reloads where `-o`-set options aren't rebuilt.

## Tests

- The `cpu` / `load` / `ram` module tests now load through the `catppuccin.tmux` entrypoint, so the escaping pass runs and their `#{E:…}` assertions exercise the render-safe round-trip. Suite passes on tmux `3.4` and `next-3.7`.
- A new **non-blocking** CI job (`continue-on-error`) builds tmux `master` and runs the suite, surfacing upstream breakage early without gating PRs for released tmux.

Manually verified on `next-3.7` with a flavor using `#1`/`#3`/`#8`/`#9`-leading colors: status modules (incl. `session`'s runtime `client_prefix` conditional), window list, and pane borders all render correct colors.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -33,3 +33,28 @@ jobs:
           bash --version
           tmux -V
           ./run_tests.sh
+
+  # Advisory canary against tmux master; non-blocking by design.
+  tmux-master:
+    name: Test (tmux master)
+    runs-on: ubuntu-24.04
+    continue-on-error: true
+
+    steps:
+      - name: Build tmux from master
+        run: |
+          sudo apt update
+          sudo apt install -y libevent-dev libncurses-dev bison autoconf automake pkg-config build-essential
+          git clone --depth 1 https://github.com/tmux/tmux.git /tmp/tmux-src
+          cd /tmp/tmux-src
+          sh autogen.sh
+          ./configure
+          make
+          sudo make install
+      - uses: actions/checkout@v6
+      - name: Run Tests
+        shell: bash
+        run: |
+          bash --version
+          tmux -V
+          ./run_tests.sh
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,14 +34,14 @@ jobs:
           tmux -V
           ./run_tests.sh
 
-  # Advisory canary against tmux master; non-blocking by design.
-  tmux-master:
-    name: Test (tmux master)
+  # Advisory canary against tmux's default branch; non-blocking by design.
+  tmux-latest:
+    name: Test (tmux latest)
     runs-on: ubuntu-24.04
     continue-on-error: true
 
     steps:
-      - name: Build tmux from master
+      - name: Build tmux from default branch
         run: |
           sudo apt update
           sudo apt install -y libevent-dev libncurses-dev bison autoconf automake pkg-config build-essential
--- a/catppuccin.tmux
+++ b/catppuccin.tmux
@@ -5,3 +5,30 @@ PLUGIN_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 
 tmux source "${PLUGIN_DIR}/catppuccin_options_tmux.conf"
 tmux source "${PLUGIN_DIR}/catppuccin_tmux.conf"
+
+# Render-safe colors.
+#
+# tmux's format_expand1 (next-3.7+) treats a leading #1-#9 as positional-argument
+# shorthand, so expanding a format at render time eats the first digit of any
+# literal #RRGGBB hex (e.g. #15232d -> 5232d). Every option below is a format that
+# tmux expands at render time (per window/pane, or via the user's #{E:...} status
+# line), so a literal hex baked into one breaks.
+#
+# Escape every #RRGGBB to ##RRGGBB; the render then unescapes it back to an intact
+# #RRGGBB. Deferred references such as #{@thm_*} carry no literal and are left
+# untouched (they resolve normally). The pass is idempotent (unescape, then
+# escape), so it stays correct if an option is not rebuilt on reload.
+_ctp_formats="window-status-format window-status-current-format pane-border-format"
+for _ctp_module in application battery clima cpu date_time directory gitmux host \
+                   kube load pomodoro_plus ram session uptime user weather; do
+  _ctp_formats="${_ctp_formats} @catppuccin_status_${_ctp_module}"
+done
+
+for _ctp_opt in ${_ctp_formats}; do
+  _ctp_val="$(tmux show -gqv "${_ctp_opt}")"
+  [ -n "${_ctp_val}" ] &&
+    tmux set -g "${_ctp_opt}" \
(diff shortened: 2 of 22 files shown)


rogeruiz:
Thanks @hoblin for this. Ill try and get this tested this weekend. I appreciate the additional CI checks for checking regressions as well but will need to scrutinize things a bit further in that area.

.github/workflows/test.yml
```diff
@@ -33,3 +33,28 @@ jobs:
...
+
+  # Advisory canary against tmux master; non-blocking by design.
+  tmux-master:
+    name: Test (tmux master)
+    runs-on: ubuntu-24.04
+    continue-on-error: true
+
+    steps:
+      - name: Build tmux from master
+        run: |
```
rogeruiz: @hoblin Can you update the descriptive references of `master` to default branch or latest, please?

```suggestion
  # Advisory canary against tmux's default branch; non-blocking by design.
  tmux-latest:
    name: Test (tmux latest)
    runs-on: ubuntu-24.04
    continue-on-error: true

    steps:
      - name: Build tmux from default branch
        run: |
```

catppuccin.tmux
```diff
@@ -5,3 +5,30 @@ PLUGIN_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
...
+# tmux's format_expand1 (next-3.7+) treats a leading #1-#9 as positional-argument
+# shorthand, so expanding a format at render time eats the first digit of any
+# literal #RRGGBB hex (e.g. #15232d -> 5232d). Every option below is a format that
+# tmux expands at render time (per window/pane, or via the user's #{E:...} status
+# line), so a literal hex baked into one breaks.
+#
+# Escape every #RRGGBB to ##RRGGBB; the render then unescapes it back to an intact
+# #RRGGBB. Deferred references such as #{@thm_*} carry no literal and are left
+# untouched (they resolve normally). The pass is idempotent (unescape, then
+# escape), so it stays correct if an option is not rebuilt on reload.
```
rogeruiz, line 20: Thanks for this descriptive comment block!

rogeruiz (changes requested):
❌ This PR needs changes ...

Just a slight change related to descriptive language in our CI. otherwise this looks to be in a good spot to integrate into `main` after those changes are made. Great work and thanks for future-proofing the project.

hoblin:
Done.

hoblin:
Hey @rogeruiz! Thanks again for the review back in June — the CI wording suggestion was a good call. I pushed that in `d402835` right after, but GitHub still shows the review as *changes requested* against the older `37d2381`, so from your side this probably looks like it's waiting on me. No rush at all — just surfacing it in case it slipped through the cracks. Still merges clean with `main`, and happy to make any other tweaks.

